PR #150, opened 2014-04-17
Create Command_imposter.java

SupItsDillon:


DUFCLiam:
@JeromSar this is not micro-management as it would add a feature we can use on people whom are imposers!

PR #401, opened 2015-01-11
Adds cooldown to /smite. Closes #400.

### LINKED ISSUE #400: Multiple Smite/Mute Prevention.
Often times when multiple admins are on people will get double or even TRIPLE smited due to one offense. My suggestion: Making a timer that would make it so someone cannot be smited within 5 seconds of being smited once. This should also apply to mute because someone who is double muted will not be muted.

Hockeyfan360:
This adds a cooldown to /smite, as suggested in the issue.


RobinGall2910:
#400

PR #418, opened 2015-02-01
Update Command_wildcard.java

vj13573:
Added 2 new commands that can't be done in a Wildcard. To prevent rogue admins.


StevenLawson:
Again, I don't think this is necessary. Admins that abuse things like this should be removed instead of blocking it in code. I can see the necessity of blocking things like "wildcard" in this command to prevent recursion, but I don't like how its implemented in this command in the first place.

Wild1145:
Why not make the new command blocker take commands like wildcard and gcmd into account when blocking? So it gets the target player and if someone is being GCMD'd then nothing simply happens, and a warning goes to the admin on screen..

PR #419, opened 2015-02-01
Update Command_gcmd.java

vj13573:


StevenLawson:
I don't think this is necessary. Admins that abuse things like this should be removed instead of blocking it in code. Also, you should make your commit messages and descriptions fit the proper guidelines.

StevenLawson:
I don't think this is necessary. Admins that abuse things like this should be removed instead of blocking it in code. Also, you should make your commit messages and descriptions fit the proper guidelines.

Wild1145:
Why not make the new command blocker take commands like wildcard and gcmd into account when blocking? So it gets the target player and if someone is being GCMD'd then nothing simply happens, and a warning goes to the admin on screen...

PR #421, opened 2015-02-04
Updated my username

Wild1145:
Due to the name change and for some reason it not ignoring case, I have had to update my username here in the developer list.


JeromSar:
All done! @Wild1145

PR #501, opened 2015-03-15
Created Command_rules.java with all rules from totalfreedom boards

looperXD:
Sometimes player wanted to view the rules but they have access to make a account on forum., But /rules give them a shorten verison of what is on forum. When a new player join they could just do /rules and quickly view all the rules and things they need to know.

Created Command_rules that has rules copied from
http://totalfreedom.boards.net/thread/21065/server-conduct-policy-players-admins

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Commands/Command_rules.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Commands/Command_rules.java
@@ -1,0 +1,63 @@
+package me.StevenLawson.TotalFreedomMod.Commands;
+
+import org.bukkit.ChatColor;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+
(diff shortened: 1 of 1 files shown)


xDestroyer217:
Or, we can just add the rules to the existing essentials command '/rules'.

Wild1145:
An issue specific to this PR should be opened prior to the PR its self and you should say that it is resolving #XYZ in the commit message.

You also will find that this will spam the chat, and may well (Unlikely) not even display all of it. I also remember this linking to another issue where multiple sub-commands would be config defined, but I cant seem to find the issue number.

JeromSar:
@Wild1145 is right, @looperXD

Wild1145:
@looperXD Are you planning to update this?

JeromSar:
@looperXD Closing for inactivity soon. Are you going to update this?

JeromSar:
I believe Essentials' /rules command is better fitted here.

PR #505, opened 2015-03-16
Resolve #436 | Block **

### LINKED ISSUE #436: Block **
It is really annoying when people use ** in commands like /msg ** and /heal **

I suggest that you block the **

zD12:
This isn't difficult, simply blocking '**' from any and all commands.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/TFM_CommandBlocker.java
+++ b/src/me/StevenLawson/TotalFreedomMod/TFM_CommandBlocker.java
@@ -117,7 +117,7 @@ public class TFM_CommandBlocker
             TFM_Util.playerMsg(sender, "Plugin-specific commands are disabled.");
             return true;
         }
-        if (command.contains("**"))
+        else if (command.contains("**"))
         {
             TFM_Util.playerMsg(sender, "'**' is prohibited.");
             return true;
--- a/src/me/StevenLawson/TotalFreedomMod/TFM_CommandBlocker.java
+++ b/src/me/StevenLawson/TotalFreedomMod/TFM_CommandBlocker.java
@@ -117,6 +117,11 @@ public class TFM_CommandBlocker
             TFM_Util.playerMsg(sender, "Plugin-specific commands are disabled.");
             return true;
         }
+        if (command.contains("**"))
+        {
+            TFM_Util.playerMsg(sender, "'**' is prohibited.");
+            return true;
+        }
 
         final String[] commandParts = command.split(" ");
         String subCommand = null;


zD12:
@BigDaddyUS it's kinda wordy and I'd rather not.  It isn't nessarily always a command, for example, '/tell ** <message>' will tell the message to all users.  However, '/eco set * 0' will reset everyone's balance, not send them/everybody a command.  That's why I just said, "'**' is prohibited."

JeromSar:
@zD12 Could you change your code to allow SuperAdmins and above to use **? That'd be nice, thanks!

JeromSar:
@zD12 Additionally, do _all_ of the commands supporting entire-server targets use a double-astrix? Are there any commands that use a single astrix instead and is there any reason why we shouldn't be blocking that instead?

Wild1145:
Would it not make sense to block anything including a singlt * as ** is still just two *'s...

zD12:
@Wild1145 and @JeromSar, I thought about that, but, factoring people use '/me hi*' for fixing typos I decided to block double.

Wild1145:
@zD12 Valid point... Not sure how we could work around that problem...

zD12:
@Wild1145 my concern/question, why would administrators need '**'?  We have '/wildcard ?'

Wild1145:
@zD12 They should not need it, neither should anyone to be fair.

JeromSar:
@zD12 Best to provide admins with everything possible in case the time comes when it *is* needed, not taking chances here. :P

Wild1145:
@JeromSar I think the point is, admins do not and should not *need* this. Wildcard will do this, and admins should be using that instead of this.

JeromSar:
@Wild1145 I don't want to block anything for admins unless absolutely necessary.

Wild1145:
@JeromSar I think not blocking it will just lead to more confusing and inconstancies with admin training. Thats the reason after all that you removed a lot of the alieses for TFM commands and changed it so admins had to execute a set second parameter instead of the original long list.

Nobody-11:
IF '**' is unblocked from admins, it would be unfair for ops, but we do have chat/wildcard admins won't be needed to use '**'

zD12:
I must agree with @Nobody-11's and @Wild1145's statements.  If we did allow double-astrixes it will only confuse administrators during training.  @JeromSar, please reconsider your modification requests.

JeromSar:
@StevenLawson what do you think?

JeromSar:
@zD12 Upon reconsideration, I think you're right, merging soon.

Wild1145:
@JeromSar Re-Open?

Wild1145:
@JeromSar Bump

Wild1145:
@JeromSar Another bump...

DUFCLiam:
@JeromSar, Bump; Needs milestone.

mayokek:
@DUFCLiam, this is made for TFM v4.3 not TFM v5.1

DUFCLiam:
Oh yeah; I shall update this if I have time.

Wild1145:

PR #527, opened 2015-03-29
Made smaller rules command

AlexFreedomMod:
i have made a smaller rules commnad.


AlexFreedomMod:
I just fixed the spelling mistakes and checked for errors.

StevenLawson:
Please test your code before you submit a pull request. This code is broken.

Brokenness aside, I believe the Essentials version of */rules* or something similar is currently being used. Even if that wasn't the case, this needs to be implemented in a way where it isn't hard coded into the source.

Wild1145:
In all honesty, if we plan to stick with essentials longer term, we are better off just modifying the essentials rules.txt / rules.yml (I cant remember which one) and using that for /rules.

AlexFreedomMod:
I think its yml, @Wild1145 you are right. The /rules on essentials would of been better.

xDestroyer217:
@Wild1145 Its rules.txt c:

PR #541, opened 2015-04-09
Solution?

looperXD:
This way when Yellow joins he will be automatically ban him. He joins and leave really fast. We won't be able to do /rangeban if he left already.
More on : http://totalfreedom.boards.net/thread/22107/banning-y3ll0w-j4ck3t-mod
Also we can't afford the server going onlinemode.
He obviously has a VPN he is not gonna stop.
This way it prevents him from logging in. Because it kicks him right after he join and bans name and IP


StevenLawson:
No. I'm not going to allow hard coded bans into TFM.

PR #552, opened 2015-04-19
Make it so telnets+ can use command blocks.

ImALuckyGuy:
First thing you may ask is "Why only telnets+ and not supers?" Well, considering command blocks effectively run the command as console, all supers would effectively be telnets if they used command blocks.

Yes, this code has been tested and works.

### DIFF
--- a/src/config.yml
+++ b/src/config.yml
@@ -198,6 +198,7 @@ console_is_senior: true
 host_sender_names:
   - rcon
   - remotebukkit
+  - '@'
 
 # Announcer
 announcer:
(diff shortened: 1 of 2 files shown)


StevenLawson:
I've written code for this before. In order for it to be secure, you have to modify the core server.

Wild1145:
It would be interesting to have some way of allowing command blocks in game but only for Telnets or really only seniors to use without having to modify Bukkit, but its making it properly secure and so that a telnet cant just place a command block and then an op come, edit it and run /saconfig add bla bla bla. The best way to do it if we want any admins (really seniors only) is to only have them allowed to be placed in the admin world.

ImALuckyGuy:
I didn't think of that. I'll close this because I just realized that and that's pretty horrible.

PR #554, opened 2015-04-19
Blocked /op causes too much spam.

IDoNotCare21:
I basically blocked op because we already have /qop which causes much less spam to op players.

### DIFF
--- a/src/config.yml
+++ b/src/config.yml
@@ -95,6 +95,7 @@ blocked_commands:
   - 'n:b:/effect:Please use /potion to set effects.'
   - 'n:b:/enderchest:_'
   - 'n:b:/spreadplayers:_'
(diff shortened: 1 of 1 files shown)


JeromSar:
@IDoNotCare21 We're not going to be doing this, sorry.

Wild1145:
Lets block /creative while we're at it.

AlexFreedomMod:
@Wild1145, Or we can block every command in TFM

Wild1145:
@AlexFreedomMod I like your style :+1:

AlexFreedomMod:
Thanks, now lets block world edit, AcidFreedom, Essentials, every plugin on totalfreedom,

TheRevalation:
@alexfreedommod Acid will murder u, 👍

PR #559, opened 2015-04-21
Create Command_owner.java

TheRevalation:
Basically to reduce spam in chat "WHO IS THE OWNER?, IS THE OWNER ONLINE?"

Needs review, I'm not sure If I forgot any imports.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Command_owner.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Command_owner.java
@@ -1,0 +1,41 @@
+package me.StevenLawson.TotalFreedomMod.Commands;
+
+import java.util.ArrayList;
+import java.util.List;
+import me.StevenLawson.TotalFreedomMod.TFM_AdminList;
+import me.StevenLawson.TotalFreedomMod.TFM_PlayerRank;
+import me.StevenLawson.TotalFreedomMod.TFM_Util;
+import org.apache.commons.lang3.StringUtils;
+import org.bukkit.ChatColor;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+
+@CommandPermissions(level = AdminLevel.ALL, source = SourceType.BOTH)
+@CommandParameters(description = "Checks if the owner is online or offline.", usage = "/<command> [-a | -i]", aliases = "who")
+public class Command_owner extends TFM_Command
+
+    @Override
+    public boolean run(CommandSender sender, Command cmd, String label, String[] args) {
+           Player slime = Bukkit.getServer().getPlayer("markbyron");
+           Player player = (Player) sender;
+        if (slime != null)
+        {
+            player.sendMessage(ChatColor.GREEN + "The owner, MarkByron, is " + ChatColor.GREEN + "online" + ChatColor.GREEN + "!");
+            return true;
+
+        }
+        else if (slime == null)
+        {
+            player.sendMessage(ChatColor.GREEN + "The owner, MarkByron, is " + ChatColor.DARK_RED + "offline" + ChatColor.GREEN + "!");
+            return true;
+        }
+
+        else
+        {
+
+        }
+        return true;
+
+    }
+    }


AlexFreedomMod:
This will be very helpful for the people that want to know if the owner is online, But some people may not know about the command. You might want to add an annoucement for it.
#559 will be helpful

TheRevalation:
I plan on doing that, if what I made works.

StevenLawson:
For this to be viable you'd want to make sure that whoever is logged in as
that user is authenticated and not an imposter. However, I'm not sure if we
really want a feature this specific in TFM...

AlexFreedomMod:
Ok

AlexFreedomMod:
@StevenLawson , yeah, There could be an imposter.

TheRevalation:
@StevenLawson them being an impostor or not wouldn't matter with the /owner command, because if the player is an impostor, he will be flagged and all will be notified, or if an admin is on, he/she will be banned for imposing mark. Anyway, it was just a suggestion, to reduce spam in chat, making it easier for everyone to see chat.

AlexFreedomMod:
@TheRevalation , some people do not look in the chat alot. And they might miss the imposter notice.
And if an admin was offline well...... They would thinkarkbyron was real

TheRevalation:
@AlexFreedomMod /list [IMP]MarkByron

 player.sendMessage(ChatColor.GREEN + "The owner, MarkByron, is " + ChatColor.GREEN + "online, but if he has [IMP] next to his name in /list, he is fake. If they are an [IMP] (imposter) please notify an admin." + ChatColor.GREEN + "!");

AlexFreedomMod:
Ok, another question... Why does it have /Owner -a and /Owner -i?

TheRevalation:
@AlexFreedomMod aliases

AlexFreedomMod:
What about it :package:

derpfacedirk:
if you use the uuid of the owner isnt it possible to prevent imposters to be shown as the owner?
you would have to know the uuid of the owner, but that cant be much of a problem
i could be overlooking something im quite new with plugins

BigDaddyUS:
You need to make sure the user is an admin. First check if the user is an imposter, then return false if the user is an imposter. If the user isn't an imposter, proceed with the command.

Wild1145:
@BigDaddyUS @derpfacedirk The way UUID works on online and offline mode servers are different. To my knowledge any offline mode servers will randomly generate a UUID on a player by player basis and use that. This means that depending on if it works on a session by session basis or not, the same username could always show as being on the same UUID (What I think happens) meaning that you have to do IP Verification either way. The best way to do this really would be to have an IS_USER_OWNER: TRUE in the superadmin.yml and if that is true and they are verified, then show they are online and real, else do not. But thats a lot more work.

JeromSar:
I'm not quite sure we want or need this command. Any thoughts, @StevenLawson?

BigDaddyUS:
@Wild1145 I don't think it has to be that difficult. Can't we just check for the string in the config.yml file? If an user is an owner, the command could list all of owners that are set in the config file then check if they're verified or not.

Wild1145:
@BigDaddyUS Okay, thats all good and well, but its just overkill. Its pulling things from a load of different places, and I just dont see there bing a gain out of it. 

The easiest way you would do it, if do an if player.getname(config thingie).isonline
bla ba bla
else
not online bla bla bla

Then run if statements within the if they are online then check if that player if verified or not.

BigDaddyUS:
@Wild1145 I was thinking something like this.
http://pastebin.com/1RHtVkeQ

Wild1145:
@BigDaddyUS While that looks like it would work, you have no verification, and thats kinda the point of the discussion for a lot of what has been said.

BigDaddyUS:
@Wild1145 Yeah I know. I wasn't writing the actually code, just setting up the idea.

src/me/StevenLawson/TotalFreedomMod/Command_owner.java
```diff
@@ -0,0 +1,41 @@
...
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+
+@CommandPermissions(level = AdminLevel.ALL, source = SourceType.BOTH)
+@CommandParameters(description = "Checks if the owner is online or offline.", usage = "/<command> [-a | -i]", aliases = "who")
+public class Command_owner extends TFM_Command
+
+    @Override
+    public boolean run(CommandSender sender, Command cmd, String label, String[] args) {
+           Player slime = Bukkit.getServer().getPlayer("markbyron");
```
Wild1145: Why have you assigned "slime" to the player? Thats just confusing, and to be honest makes it look like you ripped that code from something else pre-made. This should really be player or something similar.

The next thing, what if Mark changes his name? Your snookered then. This should really get the player by UUID if possible.

src/me/StevenLawson/TotalFreedomMod/Command_owner.java
```diff
@@ -0,0 +1,41 @@
...
+           Player player = (Player) sender;
```
Wild1145: This would be better off being "Player sender = (Player) sender;" and changing the previous line to Player player. That way the sender is refereed to as the sender, the player as the player.

src/me/StevenLawson/TotalFreedomMod/Command_owner.java
```diff
@@ -0,0 +1,41 @@
...
+            return true;
+
+        }
+        else if (slime == null)
+        {
+            player.sendMessage(ChatColor.GREEN + "The owner, MarkByron, is " + ChatColor.DARK_RED + "offline" + ChatColor.GREEN + "!");
+            return true;
+        }
+
+        else
```
Wild1145: What is this? I see no use for this else...

TheRevalation:
@Wild1145 It's my old command, I changed it up, used to be /slime, that's why it's there.

Wild1145:
@TheRevalation Probably want to change it, its not going to break anything, but its just inconsistent with the rest of the project and doesn't make a whole lot of sense here.

JeromSar:
Agreed with @StevenLawson, I don't think we really want a command this specific. I doubt the amount of spam will actually be reduced.

Issue #560, opened 2015-04-21
Telporting issues.


AlexFreedomMod:
i think /tpo and /tpohere should be blocked from ops due to Operators telporting to admins or ops tpoing other ops to them.
Also ./tptoggle should not allow others to disable someone's tptoggle, only admins should be able to do that

AlexFreedomMod:
:package: <--- contains 10530 commits after i found out there is /dtpohere and /btpohere, /ctpohere... etc

Wild1145:
Essentials plz

AlexFreedomMod:
Well... People use /tpo and /tpohere for..... Telporting people in th void, Telporting to admins without their permission, you know

AlexFreedomMod:
Closing for being inactive.

JeromSar:
@AlexFreedomMod Issues are never inactive. Only PRs are inactive when they don't get updated as requested.

Wild1145:
This issue has been superseded by #713

PR #561, opened 2015-04-21
Should fix issue #560 TPO problem

AlexFreedomMod:
I would like to fix the tpohere and tpo problem


Wild1145:
Essentials project please. Lets not start messing around and putting code in the other projects as that will end up confusing.

PR #578, opened 2015-04-29
Create a TFM-based nicking system. Fix #383

### LINKED ISSUE #383: add block for non alowed formating like &k
add block for non alowed formating like &k. to stop bad players spaming it when no admins is on.

ImALuckyGuy:
I wanted for a long time to make a nick system that prevented the forbidden chat codes from being used, but I didn't really have the knowledge until recently.

What this does it is searches for the forbidden formatting codes in your nick, then if there are those, remove them and notify the player that they have been removed.

This code has been tested and works like a charm.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Commands/Command_nick.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Commands/Command_nick.java
@@ -9,7 +9,7 @@ import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
 
 @CommandPermissions(level = AdminLevel.OP, source = SourceType.ONLY_IN_GAME)
-@CommandParameters(description = "Give yourself a nickname.", usage = "/<command> <nickname>")
+@CommandParameters(description = "Give yourself a nickname.", usage = "/<command> <nickname>", aliases = "nickname")
 public class Command_nick extends TFM_Command
 {
     @Override
--- a/src/me/StevenLawson/TotalFreedomMod/Commands/Command_nick.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Commands/Command_nick.java
@@ -1,0 +1,72 @@
+package me.StevenLawson.TotalFreedomMod.Commands;
+
+import me.StevenLawson.TotalFreedomMod.Bridge.TFM_EssentialsBridge;
+import me.StevenLawson.TotalFreedomMod.TFM_Util;
+import org.bukkit.Bukkit;
+import org.bukkit.ChatColor;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+
+@CommandPermissions(level = AdminLevel.OP, source = SourceType.ONLY_IN_GAME)
+@CommandParameters(description = "Give yourself a nickname.", usage = "/<command> <nickname>")
+public class Command_nick extends TFM_Command
+{
+    @Override
+    public boolean run(CommandSender sender, Player sender_p, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
+    {
+        if (args.length < 1)
+        {
+            playerMsg(ChatColor.RED + "You didn't provide a nickname!");
+        }
+        if (args.length > 1)
+        {
+            playerMsg(ChatColor.RED + "Too many command arguments!");
+        }
+        if ("off".equals(args[0]))
+        {
+            TFM_EssentialsBridge.setNickname(sender.getName(), null);
+            playerMsg(ChatColor.GOLD + "You no longer have a nickname.");
+            return true;
+        }        
+        final String nickPlain = ChatColor.stripColor(TFM_Util.colorize(args[0].trim()));
+        final String nickInput = args[0];
+
+        if (!nickPlain.matches("^[a-zA-Z_0-9\u00a7]+$"))
+        {
+            playerMsg(ChatColor.RED + "Your nickname contains invalid characters.");
+            return true;
+        }
+        if (nickPlain.length() < 4 || nickPlain.length() > 30)
+        {
+            playerMsg("Your nickname must be between 4 and 30 characters long.");
+            return true;
+        }
+        if (nickInput.contains("&k") || (nickInput.contains("&m") || (nickInput.contains("&n") || (nickInput.contains("&o")) || nickInput.contains("&0"))))
+        {
+            playerMsg(ChatColor.RED + "Your nickname contained forbidden chat formatting codes and as such, the forbidden codes have been removed from your nickname.");          
+        }
+        for (Player player : Bukkit.getOnlinePlayers())
+        {
+            if (player == sender_p)
+            {
+                continue;
+            }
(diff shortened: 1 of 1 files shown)


ImALuckyGuy:
Oh, and by the way, if this is merged, you'll need to add /nick to "overridden-commands" in Essentials' config.yml.

JeromSar:
@ImALuckyGuy Don't worry about that, TFM si smart and automatically overrides commans. :)

AlexFreedomMod:
Looks good. This will be very helpful

ImALuckyGuy:
Bumping this to the top. This needs review/input.

src/me/StevenLawson/TotalFreedomMod/Commands/Command_nick.java
```diff
@@ -0,0 +1,72 @@
...
+        if ("off".equals(args[0]))
+        {
+            TFM_EssentialsBridge.setNickname(sender.getName(), null);
+            playerMsg(ChatColor.GOLD + "You no longer have a nickname.");
+            return true;
+        }        
+        final String nickPlain = ChatColor.stripColor(TFM_Util.colorize(args[0].trim()));
+        final String nickInput = args[0];
+
+        if (!nickPlain.matches("^[a-zA-Z_0-9\u00a7]+$"))
```
JeromSar: Might want to have a ```public static Pattern NICKNAME_PATTERN``` to avoid reprocessing that regex every time. :)

src/me/StevenLawson/TotalFreedomMod/Commands/Command_nick.java
```diff
@@ -0,0 +1,72 @@
...
+        {
+            playerMsg(ChatColor.RED + "Your nickname contains invalid characters.");
+            return true;
+        }
+        if (nickPlain.length() < 4 || nickPlain.length() > 30)
+        {
+            playerMsg("Your nickname must be between 4 and 30 characters long.");
+            return true;
+        }
+        if (nickInput.contains("&k") || (nickInput.contains("&m") || (nickInput.contains("&n") || (nickInput.contains("&o")) || nickInput.contains("&0"))))
```
JeromSar: A method in TFM_Util would be nice. Also, this doesn't catch the actual paragraph characters. So \u00A7C would still be in Red.

src/me/StevenLawson/TotalFreedomMod/Commands/Command_nick.java
```diff
@@ -0,0 +1,72 @@
...
+            {
+                continue;
+            }
+            if (player.getName().equalsIgnoreCase(nickPlain) || ChatColor.stripColor(player.getDisplayName()).trim().equalsIgnoreCase(nickPlain))
+            {
+                playerMsg("That nickname is already in use.");
+                return true;
+            }
+        } 
+        String newNickname = nickInput.replace("&k", "").replace("&n", "").replace("&m", "").replace("&o", "").replace("&0", "").replace("&", "§");
```
JeromSar: There must be a better way to do this. And by the way, aren't nicknames with these characters already filtered out at line 43?

JeromSar:
This is a good start, but need some changes, and more review.

src/me/StevenLawson/TotalFreedomMod/Commands/Command_nick.java
```diff
@@ -0,0 +1,72 @@
...
+import me.StevenLawson.TotalFreedomMod.Bridge.TFM_EssentialsBridge;
+import me.StevenLawson.TotalFreedomMod.TFM_Util;
+import org.bukkit.Bukkit;
+import org.bukkit.ChatColor;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+
+@CommandPermissions(level = AdminLevel.OP, source = SourceType.ONLY_IN_GAME)
+@CommandParameters(description = "Give yourself a nickname.", usage = "/<command> <nickname>", aliases = "nickname")
```
JeromSar: Might want to add "off" as a second parameter too.

src/me/StevenLawson/TotalFreedomMod/Commands/Command_nick.java
```diff
@@ -0,0 +1,72 @@
...
+
+import me.StevenLawson.TotalFreedomMod.Bridge.TFM_EssentialsBridge;
+import me.StevenLawson.TotalFreedomMod.TFM_Util;
+import org.bukkit.Bukkit;
+import org.bukkit.ChatColor;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+
+@CommandPermissions(level = AdminLevel.OP, source = SourceType.ONLY_IN_GAME)
```
JeromSar: I think console should be able to remove nicknames.

src/me/StevenLawson/TotalFreedomMod/Commands/Command_nick.java
```diff
@@ -0,0 +1,72 @@
...
+public class Command_nick extends TFM_Command
+{
+    @Override
+    public boolean run(CommandSender sender, Player sender_p, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
+    {
+        if (args.length < 1)
+        {
+            playerMsg(ChatColor.RED + "You didn't provide a nickname!");
+        }
+        if (args.length > 1)
```
JeromSar: Missing the option to set nicknames for other players.

src/me/StevenLawson/TotalFreedomMod/Commands/Command_nick.java
ImALuckyGuy: Actually, no, it just tells you that the characters were removed.

JeromSar:
@ImALuckyGuy Needs some more changes, see above.

src/me/StevenLawson/TotalFreedomMod/Commands/Command_nick.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+        for (Player player : Bukkit.getOnlinePlayers())
+        {
+            if (player == sender_p)
+            {
+                continue;
+            }
+            if (player.getName().equalsIgnoreCase(nickPlain) || ChatColor.stripColor(player.getDisplayName()).trim().equalsIgnoreCase(nickPlain))
+            {
+                playerMsg("That nickname is already in use.");
```
Wild1145: Should we really stop people having identical nicknames? I do understand why you have done it, but would it not be better to check the list of online players and see if they are trying to nick themselves to be another players name, and not another persons nickname?

Wild1145:
Currently you are using Essentials as a bridge to handle nicknames, if we want to do this, we really should take ALL nickname handling from Essentials into TFM and remove Essentials nicknaming as a system all together really, else we are stretching the functionality across two plugins when it could (and should really) be contained within one.

JeromSar:
@Wild1145 is making a fair point here. Could you move all the nicking over to TFM?

ImALuckyGuy:
I'll begin working on that. The trouble is, I'd have to change several commands but I'll do as much as I know how to do.

JeromSar:
@ImALuckyGuy Any luck?

ImALuckyGuy:
I almost forgot about this. I'll resume this as soon as possible.

PR #585, opened 2015-05-01
Added another broadcast

TheRevalation:
Added another bcast

"sender - smiting player"


StevenLawson:
No thanks, I like the old message.

PR #590, opened 2015-05-04
Changing the color of Console

Scuph:
I thought since it was the same color as dev and in adminchat caused a little confusion.


AlexFreedomMod:
Hmm that would look a bit weird but i guess it will help with that.

ImALuckyGuy:
What is with you and colors? Why do you want them changed so badly?

Scuph:
@ImALuckyGuy

1. Idfk

2. Console was the same as dev, and it was confusing for me.

DaBoyTM:
Yeah nah! I don't think so, way confusing for me if you change the color, purple is fine for me

JeromSar:
To be really honest with you, i don't think this is a necessary change. Please check open PRs with the tag "Issue: Accepted" for possible contributions which will be accepted. Saves you and me a bunch of time. :)

PR #599, opened 2015-05-11
Update Command_glist.java

LegendIsAwesomes:
Adding names (Admin/Sender) to /glist ban (PlayerName)

Like my GTFO request. Clears up confusion


ImALuckyGuy:
Legend, what you modified is a change I made, and what it does is it takes the ban appeal URL from the config, instead of placing totalfreedom.boards.net there regardless of the entry.

LegendIsAwesomes:
Tested it works

JeromSar:
What are you actually trying to do here?
This doesn't add the sender name at all, it simply reverts a feature which was added several months ago. Please take some more time to study TFM before making pull requests.

PR #618, opened 2015-05-11
Fixs tons of issues at once with commands 

AlexFreedomMod:
Fixs alot of @xDestroyer217's Issues issues.

### DIFF
--- a/src/config.yml
+++ b/src/config.yml
@@ -85,6 +85,7 @@ allow:
 blocked_commands:
   # Disabled commands
   - 'n:b:/time:Server-side time changing is disabled. Please use /ptime to set your own personal time.'
+  - 'n:b:/etime:Server-Side Time changing is disabled please use /ptime instead!'
   - 'n:b:/gamemode:Use /creative and /survival to set your gamemode.'
   - 'n:b:/gamerule:_'
   - 'n:b:/ban:_'
@@ -95,8 +96,10 @@ blocked_commands:
   - 'n:b:/effect:Please use /potion to set effects.'
   - 'n:b:/enderchest:_'
   - 'n:b:/spreadplayers:_'
+  - 'n:b:/border:You are not allowed to edit the world border!.'
+  - 'n:b:/weather:Its always sunny here!'
 
-  # Superadmin commands
+  # Superadmin commands - Disabled to ops
   - 's:b:/kick:_'
   - 's:b:/socialspy:_'
   - 's:b:/kill:_'
@@ -123,6 +126,8 @@ blocked_commands:
   - 's:b:/lrbuild:_'
   - 's:b:/defaultgamemode:_'
   - 's:b:/reload:_'
+  # Soon to be blocked TFM Command i think!
+  - 's:b:/spectator:_' 
 
   # Superadmin commands - Auto-eject
   - 's:a:/stop:_'
--- a/src/config.yml
+++ b/src/config.yml
@@ -96,10 +96,10 @@ blocked_commands:
   - 'n:b:/effect:Please use /potion to set effects.'
   - 'n:b:/enderchest:_'
   - 'n:b:/spreadplayers:_'
-  - 'n:b:/border:You are not allowed to edit the world border!.'
-  - 'n:b:/weather:Its always sunny here!'
-  - 'n:b:/eweather:TotalFreedom is sunny not rainy!'
-  - 'n:b:/summon:Disabled in PR 618 Of TFM'
+  - 'n:b:/border:_'
+  - 'n:b:/weather:_'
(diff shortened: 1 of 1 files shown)


Wild1145:
First, this should be broken down (really) into separate PR's for each issue.

You also should have "Resolves #ISSUE" in the commit message where the issue is resolved, it just makes things easier.

src/config.yml
```diff
@@ -95,8 +96,11 @@ blocked_commands:
   - 'n:b:/effect:Please use /potion to set effects.'
   - 'n:b:/enderchest:_'
   - 'n:b:/spreadplayers:_'
+  - 'n:b:/border:You are not allowed to edit the world border!.'
+  - 'n:b:/weather:Its always sunny here!'
+  - 'n:b:/eweather:TotalFreedom is sunny not rainy!'
 
-  # Superadmin commands
+  # Superadmin commands - Disabled to ops
```
Wild1145: Not quite sure why we need the clarification here...

src/config.yml
```diff
@@ -123,6 +127,11 @@ blocked_commands:
   - 's:b:/lrbuild:_'
   - 's:b:/defaultgamemode:_'
   - 's:b:/reload:_'
+  # Soon to be blocked TFM Command i think!
+  - 's:b:/spectator:_'
```
Wild1145: This shouldnt be blocked. If its not already a custom TFM command then it should be created to create a spectator mode.

src/config.yml
```diff
@@ -123,6 +127,11 @@ blocked_commands:
   - 's:b:/lrbuild:_'
   - 's:b:/defaultgamemode:_'
   - 's:b:/reload:_'
+  # Soon to be blocked TFM Command i think!
+  - 's:b:/spectator:_' 
+  # Idk if i should put this here
```
Wild1145: Really just /summon would make more sense to be blocked, as for the most part its just to summon entities that you could create otherwise and would work....

AlexFreedomMod:
@Wild1145 I have fixed up all the stuff you have told me to fix

src/config.yml
```diff
@@ -85,6 +85,7 @@ allow:
 blocked_commands:
   # Disabled commands
   - 'n:b:/time:Server-side time changing is disabled. Please use /ptime to set your own personal time.'
+  - 'n:b:/etime:Server-Side Time changing is disabled please use /ptime instead!'
```
Wild1145: Would suggest using the exact same message as /time for consistency.

src/config.yml
```diff
@@ -95,6 +96,10 @@ blocked_commands:
   - 'n:b:/effect:Please use /potion to set effects.'
   - 'n:b:/enderchest:_'
   - 'n:b:/spreadplayers:_'
+  - 'n:b:/border:You are not allowed to edit the world border!.'
```
Wild1145: Minor Punctuation corrections needed at the end there.

src/config.yml
```diff
@@ -95,6 +96,10 @@ blocked_commands:
...
+  - 'n:b:/weather:Its always sunny here!'
+  - 'n:b:/eweather:TotalFreedom is sunny not rainy!'
```
Wild1145: Again, probably should use the same messages as /weather is the main command and /eweather is just an alias.

src/config.yml
```diff
@@ -95,6 +96,10 @@ blocked_commands:
...
+  - 'n:b:/summon:Disabled in PR 618 Of TFM'
```
Wild1145: Maybe something different as a message? To most that will mean nothing.

src/config.yml
JeromSar: /etime is not necessary. All aliases are blocked by default. Please remove this. :)

src/config.yml
```diff
@@ -95,6 +96,10 @@ blocked_commands:
   - 'n:b:/effect:Please use /potion to set effects.'
   - 'n:b:/enderchest:_'
   - 'n:b:/spreadplayers:_'
+  - 'n:b:/border:_'
+  - 'n:b:/weather:_'
+  - 'n:b:/eweather:_'
```
JeromSar: Another alias that doesn't need to be here.

src/config.yml
```diff
@@ -66,7 +66,7 @@ allow:
 #
 # * The second section is a letter which indicates what to do when a player executes that command.
 #   Valid actions:
-#   b - Block the command
+#   b - Block the command for everyone
```
JeromSar: No. _b_ does not block the command for everyone. It's an action, not a player-selector. _n_ is the player selector you mean.

JeromSar:
All these commands have already been blocked in the tf43 branch.

PR #647, opened 2015-05-17
Resolve #646

### LINKED ISSUE #646: Block /title
OPs may abuse such command, therefore, we should block it...

zD12:
There's not much to say about it, this resolves #646


JeromSar:
This is already added.

PR #665, opened 2015-05-24
Little fix

AlexFreedomMod:
I made a little fix in the TFM_BlockListener class.


JeromSar:
This is just a grammar change, I don't think we need this.

PR #668, opened 2015-05-26
Gj

Scuph:


JeromSar:
Eh, no.

Wild1145:
@JeromSar Why is it set out out in such a way for this command but not uniform with others?

JeromSar:
@Wild1145 This line is particularly long, in contrast to other commands.

Wild1145:
@JeromSar I mean the way that they are on separate lines, I dont believe the other commands are if I remember correctly.

PR #669, opened 2015-05-26
Create Command_nightvision

ghost:
Just a shortcut, easy to use.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Commands/Command_nightvision.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Commands/Command_nightvision.java
@@ -1,5 +1,7 @@
 package me.StevenLawson.TotalFreedomMod.Commands;
 
+import net.md_5.bungee.api.ChatColor;
+
 import org.bukkit.command.Command;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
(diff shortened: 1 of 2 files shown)


src/me/StevenLawson/TotalFreedomMod/Commands/Command_nightvision.java
```diff
@@ -0,0 +1,33 @@
+package me.StevenLawson.TotalFreedomMod.Commands;
+
+import net.md_5.bungee.api.ChatColor;
```
SlicedCaek: Use ```org.bukkit.ChatColor;```

Wild1145: Yeah, what you are using here is for a Bungee, not sure how that import even came up in all honesty because you shouldnt have a Bungee library at all.

AlexFreedomMod: Spigot has bungee in it so...... Yeah. That happens

Wild1145: Hmm Interesting. I hadnt noticed that before. But yeah, really it should be the bukkit one as thats the API we are using, not Bungee ;)

Wild1145:
Just as a general note, the formatting here is a tad haywire.

ArcticHumboldt:
I just used Eclipse's Source->Format option because my formatting was even worse, I can clean it up when I get home.

JeromSar:
@ArcticHumboldt TFM is a Netbeans project, formatting settings are only present for the NetBeans IDE, try opening TFM in there. Remember folks, NetBeans saves lives.

AlexFreedomMod:
Agreed, I remember when I used to compile TFM in eclipse xd.

ArcticHumboldt:
Noted.

OxLemonxO:
You don't need to use if (commandLabel.equalsIgnoreCase("nightvision")) {

JeromSar:
@ArcticHumboldt To be honest I don't think we really need this. What do you think @Wild1145 ?

Wild1145:
I dont think we *need* the command, I cant imagine we will have a huge amount of use as the server should always be in daylight anyway due to the way that TFM runs anyway, but some people may find it helpful if they are trying to do some sort of builds, but then again, there is a night vision potion effect... At the end of the day, if it works and its serving some purpose then I see no reason why not to add it if it might be used, but yeah...

JeromSar:
After consideration, I don't think this command is useful enough to be added to TFM, thanks for contributing anyways!

PR #672, opened 2015-05-28
Block /jail and /unjail. Fixes #671

### LINKED ISSUE #671: Block /jail and /unjail
They're unblocked atm.

ImALuckyGuy:
Today, I came on as an alt and as soon as I joined, I was jailed and the jailer refused to let me out even after I told him I was an admin. He only believed me after I went ingame on my normal account and noobed him. Superadmins won't be needing this command either as we have /cage.

### DIFF
--- a/src/config.yml
+++ b/src/config.yml
@@ -97,6 +97,7 @@ blocked_commands:
   - 'n:b:/spreadplayers:_'
   - 'n:b:/execute:_'
   - 'n:b:/jail:_'
+  - 'n:b:/unjail:_'
 
   # Superadmin commands
   - 's:b:/kick:_'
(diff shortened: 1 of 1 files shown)


JeromSar:
@ImALuckyGuy Maybe admin-only?

Wild1145:
Why not just do this in Essentials?

Scuph:
@Wild1145 because it's 10x easier

ImALuckyGuy:
I've done this in TF-Essentials now, see https://github.com/TotalFreedom/TF-Essentials/pull/5

Wild1145:
@Scuph Just because its easier doesnt make it the best way to do things. If we always did things the easy way we would just permban anyone who breaks the rules and not need TFM.

PR #680, opened 2015-06-02
Change the force-IP to a contains statement to fix Forge issue. Fixes #493

### LINKED ISSUE #493: Forge Mod Loader workaround
[01:37:18] [Server thread/INFO]: [TotalFreedomMod] Hostname: play.totalfreedom.meFML:28965]

Forge Mod Loader appends FML to the hostname in the log, which is  what prevents a player using at least certain versions FML from joining TotalFreedom.  I dont get how they are even managing to establish a connection though with the suffix. I would think that it would fail to resolve the domain with meFML.

Anyways, this is why some FML users are getting kicked for not joining with the correct IP/domain.

ImALuckyGuy:
I had to change some things to toLowerCase() to prevent capitalization issues, but it does no harm.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/TFM_ServerInterface.java
+++ b/src/me/StevenLawson/TotalFreedomMod/TFM_ServerInterface.java
@@ -112,11 +112,11 @@ public class TFM_ServerInterface
         // Check force-IP match
         if (TFM_ConfigEntry.FORCE_IP_ENABLED.getBoolean())
         {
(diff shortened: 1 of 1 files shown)


Wild1145:
I cannot see that this will fix anything, Forge Mod Loader adds FML into the hostname, so all you're doing here is reverting one commit and tweaking the reversion.

ImALuckyGuy:
No, what it does is it checks if the hostname _contains_ play.totalfreedom.me, so regardless if FML is there or not, it will allow the player to join.

JeromSar:
@ImALuckyGuy Looking at #493, I think @ChargedCreeper has found a better solution here. Thanks for contributin anyway!

PR #687, opened 2015-06-03
Fix for Forge

### LINKED ISSUE #493: Forge Mod Loader workaround

ChargedCreeper:
Dug a bit deeper and FML isnt the full string of what is injected. There are also null characters that should be removed. I found this by using a binary editor on a LAN server log file.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/TFM_ServerInterface.java
+++ b/src/me/StevenLawson/TotalFreedomMod/TFM_ServerInterface.java
@@ -112,7 +112,7 @@ public class TFM_ServerInterface
         // Check force-IP match
         if (TFM_ConfigEntry.FORCE_IP_ENABLED.getBoolean())
         {
-            final String hostname = event.getHostname().replace("FML", null); // Forge fix - https://github.com/TotalFreedom/TotalFreedomMod/issues/493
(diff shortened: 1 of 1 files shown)


JeromSar:
@ChargedCreeper Oh dear, null characters?? Did you verify this works?

ChargedCreeper:
@JeromSar Yes I did verify it works. Compiled TFM locally, enabled force IP and was still able to connect when I compiled with this change.

JeromSar:
@ChargedCreeper Alright, thanks for the fix! :) Merging soon.

JeromSar:
Merged in https://github.com/TotalFreedom/TotalFreedomMod/commit/1ed1b352f301e373b64769f79909c9cfec2da5be Thanks for contributing!

Adaptified:
This was already merged @JeromSar.

DUFCLiam:
@JeromSar Remove the "PR: Needs CLA" and add a new label "PR: Merged"

PR #699, opened 2015-06-10
Closes #698

### LINKED ISSUE #698: Set Op Sign Exploit
There is a new exploit, it crashes players.

This needs to be fixed.

Scuph:
My apologies, I thought of this so wrong >.<


ImALuckyGuy:
Isn't this an awful idea? Blocking signs for ops is a massive removal of freedom.

Hockeyfan360:
Yes, you need to get the block, and get the text on line 1/line2, etc. If it equals that, then you cancel the event. Otherwise, this is a pitiful idea.

ImALuckyGuy:
It isn't even player.setOp(null) that does this. Your facts are wrong.

Scuph:
Read the initial comment above, I messed up!

PR #702, opened 2015-06-11
Add smite reasons

ImScoop:
Back from my old issues, Here is my PR for it.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Commands/Command_smite.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Commands/Command_smite.java
@@ -30,7 +30,7 @@ public class Command_smite extends TFM_Command
             return true;
         }
 
-        smite(player);
+        smite(player, reason);
 
         return true;
     }
@@ -38,7 +38,7 @@ public class Command_smite extends TFM_Command
     final String smiteReason = StringUtils.join(args, 1, args.length), " ");
      
 
-    public static void smite(final Player player)
+    public static void smite(final Player player, String reason)
     {
 
         if (smiteReason == null)
--- a/src/me/StevenLawson/TotalFreedomMod/Commands/Command_smite.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Commands/Command_smite.java
@@ -10,7 +10,7 @@ import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
 
 @CommandPermissions(level = AdminLevel.SUPER, source = SourceType.BOTH)
-@CommandParameters(description = "Someone being a little bitch? Smite them down...", usage = "/<command> [playername] [reason]")
+@CommandParameters(description = "Someone being a little bitch? Smite them down...", usage = "/<command> [playername] [reason:optional]")
 public class Command_smite extends TFM_Command
 {
     @Override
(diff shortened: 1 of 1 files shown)


Wild1145:
I think (if I remember) that @JeromSar has always wanted reasons to be optional, I personally think they should be a requirement, but if I remember correctly I think that it was raised before to be an optional setting.

Scuph:
@Wild115 So the make the minimum args length one?

src/me/StevenLawson/TotalFreedomMod/Commands/Command_smite.java
```diff
@@ -4,19 +4,21 @@
...
 @CommandPermissions(level = AdminLevel.SUPER, source = SourceType.BOTH)
-@CommandParameters(description = "Someone being a little bitch? Smite them down...", usage = "/<command> [playername]")
+@CommandParameters(description = "Someone being a little bitch? Smite them down...", usage = "/<command> [playername] [reason]")
 public class Command_smite extends TFM_Command
 {
     @Override
     public boolean run(CommandSender sender, Player sender_p, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
     {
-        if (args.length != 1)
+        if (args.length != 2)
```
JeromSar: Reason should be optional.

src/me/StevenLawson/TotalFreedomMod/Commands/Command_smite.java
```diff
@@ -33,10 +35,12 @@ public boolean run(CommandSender sender, Player sender_p, Command cmd, String co
 
         return true;
     }
+    
+    String smiteReason = StringUtils.join(ArrayUtils.subarray(args, 1, args.length), " ");
```
JeromSar: StringUtils.join can do this without requiring ArrayUtils. Check the APIdocs

src/me/StevenLawson/TotalFreedomMod/Commands/Command_smite.java
```diff
@@ -33,10 +35,12 @@ public boolean run(CommandSender sender, Player sender_p, Command cmd, String co
...
 
     public static void smite(final Player player)
     {
-        TFM_Util.bcastMsg(player.getName() + " has been a naughty, naughty boy.", ChatColor.RED);
+        TFM_Util.bcastMsg(player.getName() + " has been a naughty, naughty boy.\nThey have thus been smitten/nReason - " + smiteReason, ChatColor.RED);
```
JeromSar: @Scuph Broadcasts shouldn't have newlines. Additionally "They have thus been smitten" is unnecessary. If there's a reason present, it should be on the same line.

Scuph:
I fixed some things, but the stringutils I will look into

src/me/StevenLawson/TotalFreedomMod/Commands/Command_smite.java
```diff
@@ -33,10 +35,12 @@ public boolean run(CommandSender sender, Player sender_p, Command cmd, String co
 
         return true;
     }
+    
+    String smiteReason = StringUtils.join(args, 1, args.length), " ");
```
JeromSar: This doesn't work if there's no reason. Test your code!!

Scuph:
@Jeromsar I said id look later

Scuph:
@JeromSar done.

Scuph:
@JeromSar You keep marking Inactivity on my PRs, while this one was tested and works, please review what you are doing.

Adaptified:
Still sitting here :)

Adaptified:
@JeromSar Tried a more simpler method.

JeromSar:
@Adaptified There's still some issues with this PR:
* Reasons aren't optional
* Nonstandard usage format
* It does not compile, let alone work.

PR #706, opened 2015-06-15
Block Replacenear

SlicedCaek:


SlicedCaek:
@JeromSar, I have tested it on my LAN Server. Worked!

WickedGamingUK:
I'm against this idea. Takes away players freedom. Instead, a limit needs to be introduced so the server doesn't crash because of it.

Scuph:
@WickedGamingUK I agree.

AlexFreedomMod:
@WickedGamingUK Agreeded.

OxLemonxO:
I  used to use removenear when I was an op.
I also think that a limit should be there instead of just blocking it.

Wild1145:
I think it should be blocked, there is no valid reason to use it other than destroying other player builds... I can now easily go to another players build, run a remove near command and no trace I did it...

Scuph:
@Wild1145 it can easily be used for roleplay where in some occasion a certain block type might need to change instantly.

SlicedCaek:
I am on @Wild1145  side... Doesn't matter what limit you got... you can still remove a building (Spawn etc) and escape with it (No trace)

WickedGamingUK:
So while we're at it should we block WorldEdit completely for the same reason? No, that's a stupid idea, exactly the same as blocking this command.

Wild1145:
@WickedGamingUK The difference between blocking worldedit completely and this, is that this has no non-vicious uses that I can think of, whereas a lot of the other worldedit commands have valid useful parts to them, this just removes anything nearby without leaving a trace, even limits wouldnt help here.

JeromSar:
@WickedGamingUK @Wild1145 @Scuph @OxLemonxO @AlexFreedomMod After careful consideration I think we don't want this command blocked. Although it may be used to grief, it's also a vital component of WorldEdit, and blocking it would thus impose more restrictions on whatever's left of the freedom on TF.

We can however modify WorldEdit-TF so that this command follows the regular limits WorldEdit has set . This should reduce the amount of crashing of TF.

@SlicedCaek All in all, thanks for contributing, but I don't think this addition is right for TFM.

PR #730, opened 2015-06-28
Block /regen and /replacenear

LegendIsAwesomes:
Well, All these commands are used for is crashing the server.

Why not just block them so only SuperAdmin+ can use these commands.

### DIFF
--- a/src/config.yml
+++ b/src/config.yml
@@ -134,6 +134,8 @@ blocked_commands:
   - 's:b:/green:_'
   - 'n:b:/gamerule:_'
   - 'n:b:/togglejail:_'
+  - 'n:b:/regen:_'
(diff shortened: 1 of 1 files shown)


Wild1145:
Resolves?

ScuphGamingUK:
@Wild115 These people don't seem to read the contributing guidelines. I didn't after my first, 2 PRs? But legend always does it, you need to make PRs that are valid and solve issues. If they solve an issue, put what issue it solves. If it is just a contributing PR, make a detailed description.

JeromSar:
@LegendIsAwesomes If these commands pose a problem, they should be modified in TF-WorldEdit, not simply blocked.

Issue #716, opened 2015-06-21
Muted players can use /emsg


AlexFreedomMod:
Muted players can use /emsg, /er, /eme  since they are not blocked for muted players.

jjkatz613:
It is an easy fix in the code. There is a place for it.

AlexFreedomMod:
mhm i just wanted to point it out

PattyToTheWatty:
Well, for the mean time just /mute or /stfu them then block all ocmmands. ez

looperXD:
They also could do /w /whisper /ew /ewhisper I believe

Scuph:
@PattyToTheWatty said:
"Well, for the mean time just /mute or /stfu them then block all ocmmands. ez"

Or just make a pull request adding them to the blocked commands while muted section.

LegendIsAwesomes:
Well, This isn't exactly a bug. I'll make a PR sorting this out.

UltimaTheHawke:
Apologies for the kind-of bump, but this can also be used with /action as an alternative to /me. Mind blocking that when muted?

PR #732, opened 2015-06-28
Blocking muted from using alternate speaking commands.

LegendIsAwesomes:
https://github.com/TotalFreedom/TotalFreedomMod/issues/716

Just blocks them from doing the commands mentioned in the issue.

I think this is correct. Not too sure.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_PlayerListener.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_PlayerListener.java
@@ -61,7 +61,7 @@ import org.bukkit.util.Vector;
 
 public class TFM_PlayerListener implements Listener
 {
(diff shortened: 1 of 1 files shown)


Wild1145:
Would have been more sensible to either add them all at the end of the list or add the alternate commands next to the current ones personally.

Adaptified:
Might as well remove say, because it's admin only.

looperXD:
May you add /action? It works same as /me

LegendIsAwesomes:
Added /action. @JeromSar

JeromSar:
@LegendIsAwesomes This is good, but needs a different implementation. Instead of blocking all the aliases, please change it to automatically block all aliases by resolving each command name to its associated `Command` instance. That way, aliases are automatically blocked.

PR #733, opened 2015-06-28
Highlight OP's commands with a color.

LegendIsAwesomes:
Title says it all.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_PlayerListener.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_PlayerListener.java
@@ -729,7 +729,7 @@ public class TFM_PlayerListener implements Listener
             {
                 if (TFM_AdminList.isSuperAdmin(pl) && TFM_PlayerData.getPlayerData(pl).cmdspyEnabled())
                 {
-                    TFM_Util.playerMsg(pl, player.getName() + ": " + command);
+                    TFM_Util.playerMsg(pl, ChatColor.GREEN + player.getName() + ": " + command);
                 }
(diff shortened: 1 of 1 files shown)


ScuphGamingUK:
No, it creates chat spam.

LegendIsAwesomes:
Oops @JeromSar I accidentally closed it. Can you open it.

looperXD:
I think there is  a re-open button next to the comment bottom for you @legendisawesome

Wild1145:
They are grey to stop the messages looking like chat messages...

LegendIsAwesomes:
@Wild1145 Can you not change it to a color light purple instead?

Wild1145:
@LegendIsAwesomes You could, but again, it was set like this so its not distracting...

LegendIsAwesomes:
@Wild1145 I know what you're talking about but i find grey way too hard to look at when searching for a cmd a op did.

Wild1145:
@LegendIsAwesomes That is true... Another colour would be useful, its just finding which one... Is it ChatColor.GREY or ChatColor.DARK_GREY? If its GREY maybe we should try DARK_GREY (As I think it has one of those)

Hockeyfan360:
Bold the text?

LegendIsAwesomes:
@wild1145 DARK GREY would work better if it was implemented, I'll make a commit later.

Hockeyfan360:
I'm going to be brutally honest, this is absolutely useless and would be extremely distracting.

JeromSar:
@LegendIsAwesomes This would only make OP commands harder to read. I don't think we want this, sorry.

PR #776, opened 2015-08-10
Don't remove italics when using /nickclean.

Adaptified:
Currently italics are allowed; and nickclean wipes any user's nickname if it contains such characters.
See http://totalfreedom.boards.net/thread/26577/allow-italics.


GangSterFlexx:
Don't make a PR unless it have been approved and also you will required to make a issue.

Wild1145:
@ImScoop What issue does this link to?

PR #779, opened 2015-08-10
Add onlinemode on/off notification - closes #778

### LINKED ISSUE #778: MotD Online/Offline mode Notif
I was thinking that since the server sometimes goes between the two there could be a notification of some sort saying that it's in online mode.

for example, an unfilled character would represent offlinemode, and a filled one would represent onlinemode (e.g. ○), while still being colored, like normal.

ImALuckyGuy:
What this does is if onlinemode is on, it adds a full circle "•" at the beginning of the MOTD. If it is off, it adds an empty circle at the end "○".

This code has been tested and works exactly as it should.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_ServerListener.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_ServerListener.java
@@ -51,6 +51,8 @@ public class TFM_ServerListener implements Listener
         // Colorful MOTD
 
         final StringBuilder motd = new StringBuilder();
+        
+        motd.append(Bukkit.getServer().getOnlineMode() ? "•" : "○");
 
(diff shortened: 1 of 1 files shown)


GangSterFlexx:
Bump - LulzCraft are using this code and looks like it is working.

Wild1145:
Yeah, this should work. Not sure if the idea of bullet points really is ideal, but it does work.

JeromSar:
I think @Wild1145 has a fair point. Could you make a different implementation?

JeromSar:
@ImALuckyGuy Additionally, could you allow this marker to be configurable, just as `%mcversion%`?

Wild1145:
@JeromSar Re-Open?

DUFCLiam:

JeromSar:
Original issue was closed.

PR #787, opened 2015-08-19
Add a Telnet Clan. Admin rank.

FenixCraft:
We should add a Telnet Clan. Admin rank that can be toggled in superadmin.yml just like for STA and SrA. It'll be easier to compare whos normal STA and whos TCA because lots of people get confused. The colour for it can be &a and I have no clue anything else to say. :+1: 

Thanks,
~Fenix


JeromSar:
No.

PR #788, opened 2015-08-20
Persistently store commandspy

### LINKED ISSUE #785: Make cmdspy remember if an admin toggled itself
So instead of just doing /cmdspy over and over again whenever the server restarts or you log into it, what if we made it so that in the playerdata it will remember if you toggled commandspy previously or not and if you toggled it on it would stay on until you toggled it off.

tylerhyperHD:
Creates an adminlist entry that saves if the admin toggled cmdspy.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Commands/Command_rank.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Commands/Command_rank.java
@@ -10,6 +10,7 @@ import org.bukkit.entity.Player;
 @CommandParameters(description = "Shows your rank.", usage = "/<command>")
 public class Command_rank extends TFM_Command
 {
+
     @Override
     public boolean run(CommandSender sender, Player sender_p, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
     {
--- a/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_PlayerListener.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_PlayerListener.java
@@ -723,7 +723,7 @@ public class TFM_PlayerListener implements Listener
             // CommandBlocker handles messages and broadcasts
             event.setCancelled(true);
         }
-        
+
         if (!TFM_AdminList.isSuperAdmin(player))
         {
             for (Player pl : Bukkit.getOnlinePlayers())
--- a/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_PlayerListener.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_PlayerListener.java
@@ -724,14 +724,11 @@ public class TFM_PlayerListener implements Listener
             event.setCancelled(true);
         }
         
-        // Grab adminlist values
-        TFM_Admin admin = TFM_AdminList.getEntry(player);
-        
         if (!TFM_AdminList.isSuperAdmin(player))
         {
             for (Player pl : Bukkit.getOnlinePlayers())
             {
-                if (TFM_AdminList.isSuperAdmin(pl) && admin.cmdSpyEnabled())
+                if (TFM_AdminList.isSuperAdmin(pl) && TFM_AdminList.getEntry(pl).cmdSpyEnabled())
                 {
                     TFM_Util.playerMsg(pl, player.getName() + ": " + command);
                 }
--- a/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_PlayerListener.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_PlayerListener.java
@@ -9,6 +9,7 @@ import java.util.Random;
 import java.util.regex.Pattern;
 import me.StevenLawson.TotalFreedomMod.Commands.Command_landmine;
 import me.StevenLawson.TotalFreedomMod.Config.TFM_ConfigEntry;
+import me.StevenLawson.TotalFreedomMod.TFM_Admin;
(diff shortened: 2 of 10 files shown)


Wild1145:
That is all good and well, but the first time that is used, its going to cause errors after errors, because there will be a missing entry for every admin on the superadmin.yml...

tylerhyperHD:
Yeah I did think long and hard about resolving that, but there was no other feasible option to make it so that it went around doing that.

JeromSar:
@Wild1145 Actually, I don't believe that is the case, It seems like that is already taken care of.

@tylerhyperHD This looks really good, I'll have to try it out, but it seems like you put real effort into designing this. Thanks for contributing!

Wild1145:
@JeromSar Are we not going to get an NPE from there being no true / false option in the superadmin.yml... Or did I miss something?

JeromSar:
@Wild1145 No, because of two things:
* `boolean` type variables cannot have null values in Java.
* `getBoolean()` is called with a second argument, which defines the default value.

tylerhyperHD:
Tested it and it has some errors in it, and I don't seem to know where they are.

src/me/StevenLawson/TotalFreedomMod/Listener/TFM_PlayerListener.java
```diff
@@ -722,12 +723,15 @@ public void onPlayerCommandPreprocess(PlayerCommandPreprocessEvent event)
...
+        
+        // Grab adminlist values
+        TFM_Admin admin = TFM_AdminList.getEntry(player);
+        
         if (!TFM_AdminList.isSuperAdmin(player))
         {
             for (Player pl : Bukkit.getOnlinePlayers())
             {
-                if (TFM_AdminList.isSuperAdmin(pl) && TFM_PlayerData.getPlayerData(pl).cmdspyEnabled())
+                if (TFM_AdminList.isSuperAdmin(pl) && admin.cmdSpyEnabled())
```
JeromSar: Incorrect reference to adminlist entry.

src/me/StevenLawson/TotalFreedomMod/TFM_Admin.java
```diff
@@ -46,6 +48,7 @@ public TFM_Admin(UUID uuid, ConfigurationSection section)
         this.isSeniorAdmin = section.getBoolean("is_senior_admin", false);
         this.isTelnetAdmin = section.getBoolean("is_telnet_admin", false);
         this.consoleAliases = section.getStringList("console_aliases");
+        this.cmdSpyEnabled = section.getBoolean("cmdspy_is_on", false);
```
JeromSar: How about just `cmdspy`?

src/me/StevenLawson/TotalFreedomMod/Listener/TFM_PlayerListener.java
tylerhyperHD: How should I write it then if I were to do this?

AlexFreedomMod:
I don't see what Admin stands for in the code. (It doesn't mean Anything right now)

src/me/StevenLawson/TotalFreedomMod/Listener/TFM_PlayerListener.java
JeromSar: @tylerhyperHD Follow the logic:
* `admin` refers to the entry for the player executing the command, not the looped player.
* `admin` may be null.

tylerhyperHD:
I have committed and fixed my issues.

src/superadmin.yml
```diff
@@ -8,6 +8,7 @@ admins:
   f9a1982e-252e-4ed3-92ed-52b0506a39c9:
     last_login_name: Prozza
     is_activated: true
+    cmdspy_is_on: false
```
AlexFreedomMod: As @JeromSar suggested it should  be on by default so instead of having cmdspy is on on you should make it cmdspy is off in the configuration

src/me/StevenLawson/TotalFreedomMod/TFM_Admin.java
tylerhyperHD: The reason I made it false it to prevent everyone from getting commandspy when they first join if the admin list is wiped.

src/superadmin.yml
tylerhyperHD: I dun goofed and forgot to change the config string but:
The reason I made it false it to prevent everyone from getting commandspy when they first join if the admin list is wiped.

JeromSar:
@tylerhyperHD This looks good. I'll need to test this, and hopefully, if there are no issues, this PR will be merged soon. :)

Wild1145:
@JeromSar Re-Open?

Adaptified:
This PR looks really good, so I hope he signs it.

DUFCLiam:
And CLA will likely never be signed.

Wild1145:
The branch is invalid from where it is being merged from and the destination branch is incorrectly.

src/me/StevenLawson/TotalFreedomMod/Commands/Command_rank.java
```diff
@@ -10,6 +10,7 @@
 @CommandParameters(description = "Shows your rank.", usage = "/<command>")
 public class Command_rank extends TFM_Command
 {
+
```
Wild1145, line 13: Not sure why you have added an additional line here.

Wild1145 (changes requested):
Please address the random additional line, and also address the conflicts that will not allow this PR to merge with the master branch.

PR #816, opened 2015-09-08
Added /spectator command

TheWadeSmith:
Interesting command. I don't see how people can break the rules with it. It will also help admins get around easier.


Wild1145:
Trolling would be my only concern. Personally I would limit it to admins only and have this as an admin tool.

CrafterSmith12:
How would they troll with it exactly?

JeromSar:
@CrafterSmith12 I don't understand why players would want to be in spectator mode. Could you enlighten me on that? :)

AlexFreedomMod:
Ops can grief with it I suggest making it for admins only kinda like a better invis potion

CrafterSmith12:
@JeromSar It's manly for admins, however, I don't see an issue with players having it. Please correct me of I'm wrong.

JeromSar:
If I'm not mistaken, players in spectator mode are invisible to other players. Thus, I think that adding this command will only add chaos, and wouldn't benefit TF's community.

Wild1145:
@JeromSar You are invisible to all players not in Spectator mode. Thus why I would suggest it be strictly for admins only, and not OP's, as while in spectator mode you cannot interact with the world or other players, it would be a very spy orientated tool and could be used to troll other players.

CrafterSmith12:
@JeromSar

src/me/StevenLawson/TotalFreedomMod/Commands/Command_spectator.java
```diff
@@ -0,0 +1,35 @@
...
+    public boolean run(CommandSender sender, Player sender_p, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
+    {
+        Player player;
+        if (args.length == 0)
+        {
+            player = sender_p;
+        }
+        else
+        {
+            if (!(senderIsConsole || TFM_AdminList.isSuperAdmin(sender)))
```
JeromSar: Only supers can execute this command, this is redundant.

Wild1145:
@JeromSar Re-Open?

DUFCLiam:

PR #818, opened 2015-09-08
Made /opall -a and /opall -sp

CrafterSmith12:
-a = adventure gamemode
-sp = spectator gamemode


DarkGamingDronze:
Spectator gamemode would make a total choas because there whould be 45 ghosts flying around, and it might be abused by admins.

JeromSar:
After some review, I don't think we want/need this. Thanks for contributing!

PR #852, opened 2015-10-13
Changing impostors gamemode.

CrafterSmith12:
Players who pose as admins can still grief around them even though they're frozen. I made it so that if they join as an impostor, they're set to adventure mode instead of the default survival.


xDestroyer217:
Well, this is correct, adventure doesn't allow you to grief at all.

JeromSar:
@CrafterSmith12 Could you make it so freezing a player always sets them to adventure mode? When a player is unfrozen, change them back to survival. That way, we're sure that no frozen player can grief.

JeromSar:

PR #855, opened 2015-10-14
Blocks snow and green. Resolves #854.

### LINKED ISSUE #854: Block /green and //snow
This is not even used at all by OPs, these just allow the players to crash the server by doing /green 9001 and such. Same as with snow. I mean it's a great use for people who tend to use it the right way, so I recommend to allow to people to use this when they're admin.

RobinGall2910:
#854


JeromSar:
Can snow both be used as '/snow' and '//snow'? Is one an alias over another. If this is the case, then one of the two is redundant.

JeromSar:

PR #861, opened 2015-10-18
Resolves #860 - Blockcmd message changed

### LINKED ISSUE #860: Change /blockcmd Message
For a while now, it's just been bugging me, the message "Blocking all commands", it just doesn't sound right. Instead, I was thinking it could be "Blocking all command input". While they should get the message in general, this is much more understandable and better looking.

Rufure:
#860

Changes "block commands" to "command input".


JeromSar:
After some consideration, I think it's fine the way it is currently.

PR #898, opened 2015-11-14
Grammar mistake

CrafterSmith12:


Wild1145:
I may be getting this muddled up with web based languages, but cant the ' 's cause issues with formatting?

Obi2:
I don't think the original message was erroneous - "everyone's" should be correct.
https://english.stackexchange.com/questions/6058/should-i-use-everyones-everyones-or-everyones

OutOfYourMind:
I agree with Obi2, the apostrophe is to show possession, you possess the creative gamemode.

JeromSar:
After some review, @Obi2  turns out to be right, closing.

PR #909, opened 2015-11-29
Make /opall -s seniors only. Resolves #897

### LINKED ISSUE #897: Block /opall -s to seniors
I don't think -s is needed as most of people always be on creative.

JacobMathBoy:
I think that the -s on /opall should be for seniors only, but I agree:
> Why would we ever *remove* functionality from TFM?


JacobMathBoy:
I'm sorry, please deny this one. I forgot where the NO_PERMISSION message was.

PR #906, opened 2015-11-28
Put drops in inventory for survival mode players.

LegendIsAwesomes:
Should work in TFM, Resolves #889

PR #910, opened 2015-11-29
Resolves #889

### LINKED ISSUE #889: Survival mode enhancement
When someone is in survival mode and they break a block, instead of the item dropping on the ground (and thus being auto-removed) it would go directly into their inventory. Why this would be useful: ops could set up and play things like Skyblock on the server a lot easier. As it is currently, survival mode play on the server is not practical because of how TFM handles drops, and this would improve that quite a bit without introducing entity lag issues.

LegendIsAwesomes:
Resolves #889 (Sorry, I created a new one by accident, original one can be found here: https://github.com/TotalFreedom/TotalFreedomMod/pull/906 ) @JeromSar

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_BlockListener.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Listener/TFM_BlockListener.java
@@ -10,6 +10,7 @@ import me.StevenLawson.TotalFreedomMod.TFM_RollbackManager;
 import me.StevenLawson.TotalFreedomMod.TFM_Util;
 import me.StevenLawson.TotalFreedomMod.TotalFreedomMod;
 import org.bukkit.ChatColor;
+import org.bukkit.GameMode;
 import org.bukkit.Location;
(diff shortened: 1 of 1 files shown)


JeromSar:
@LegendIsAwesomes Please append your changes to https://github.com/TotalFreedom/TotalFreedomMod/pull/906

src/me/StevenLawson/TotalFreedomMod/Listener/TFM_BlockListener.java
```diff
@@ -111,6 +112,17 @@ else if (playerLocation.distanceSquared(location) > (nukeMonitorRange * nukeMoni
...
             }
         }
+        
+        if (player.getGameMode().equals(GameMode.SURVIVAL))
+        {
+            for (ItemStack block : event.getBlock().getDrops())
+            {
+                event.getBlock();
+                player.getInventory().addItem(block);
+                event.setCancelled(true);
```
DUFCLiam: Why the cancel?

PR #912, opened 2015-11-30
Resolves #897 - Block /opall -s to seniors

### LINKED ISSUE #897: Block /opall -s to seniors

AvalancheYT:
I have tested it on my LAN server even I checked as Senior.

Everything is working.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Commands/Command_opall.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Commands/Command_opall.java
@@ -24,14 +24,17 @@ public class Command_opall extends TFM_Command
                 doSetGamemode = true;
                 targetGamemode = GameMode.CREATIVE;
             }
-            else if (TFM_AdminList.isSeniorAdmin(sender))
+            else if (args[0].equals("-s"))
             {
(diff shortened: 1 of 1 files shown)


JeromSar:
@Verdon-Online The associated ticket has not yet been accepted.

src/me/StevenLawson/TotalFreedomMod/Commands/Command_opall.java
```diff
@@ -24,12 +24,15 @@ public boolean run(CommandSender sender, Player sender_p, Command cmd, String co
                 doSetGamemode = true;
                 targetGamemode = GameMode.CREATIVE;
             }
-            else if (args[0].equals("-s"))
```
Wild1145: Not sure why you removed this to replace it with a permissions check. Surly you would be better off leaving this here and performing the permissions check within this IF statement. That way if they are not senior you can easily throw a no perms error, where here you are not able / have done this... Just makes more work for yourself this way around.

JeromSar:
Associated issue has been closed.

PR #926, opened 2015-12-13
Update smite

DUFCLiam:
Update boy to "person" and tell the user who they where smitted by.


JeromSar:
@DUFCLiam This use of grammar is intentional.

PR #934, opened 2015-12-16
I Got An Idea To Add Reporter Rank To TotalFreedomMod Please Read It all

LightU:
I Got An Idea To Make New Ranks Head Reporter And Reporter please read it all .
Head Reporter Would Have Permission To Use Log Stick . And Also Use /Rollback, /jail and /smite
Reporter Would Use Log Stick and /smite But Not /rollback and /jail

That Idea For Reporter Was Made Because That Would Be Useful like A Little Police That Would Find Out Griefers Easily And Report Them To Admins If There Not admin online They would Use command /jail
To Stop Players From Griefing . if there's not Head Reporter ON Only Reporter and no admins Reporter Could Just /smite Griefer My nick in game Play.totalFreedom.me Is LightUk I would like to Become Head Reporter . I Am Trying To Stop Griefers in server by /report but when no admins on it sucks and its also harder to find out griefers . Thanks for reading my idea :D I believe You Add This update To TotalFreeDomMod :DD
Sorry For my hand writing I don't really come from England So my English isn't really good :DD


JeromSar:
@LightU You have no idea what you're doing.

PR #938, opened 2015-12-19
A good Update to TFM!

AlphaGaming:
Updating the Server: to TotalFreedom:

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Commands/Command_say.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Commands/Command_say.java
@@ -38,7 +38,7 @@ public class Command_say extends TFM_Command
             }
         }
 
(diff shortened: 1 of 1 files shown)


PacksGamingHD:
its fine the way it is

Adamx120:
Unnecesarry

OnlineRepz:
How that even a good update... Stop making PRs without issues (accepted)

JeromSar:
No thank you.

Wild1145:
It would however be good to change it to dynamically update based on the config value defined here: https://github.com/TotalFreedom/TotalFreedomMod/blob/master/src/config.yml#L9

JeromSar:
@Wild1145 No I think this should stay as "Server" since that is the Bukkit/Minecraft default.

Wild1145:
@JeromSar Eh, I think either works. If it is going to be changed then I dont think we need it to be something static like "TotalFreedom" though, thus why I suggested the dynamic update.

PR #942, opened 2015-12-19
there are 101 commands :P

ghost:


OnlineRepz:
No thanks

rsajeey:
Over 85 means above 85+ not static 85.

JeromSar:
@rsajeey is right.

PR #949, opened 2015-12-19
Adding /clearparticles

Adamx120:
I think it is an useful command cause PlayerParticles plugin is lagging TotalFreedom
and i think it is easier than typing ./wildcard gcmd ? pp clear.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Commands/Command_clearparticles.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Commands/Command_clearparticles.java
@@ -12,12 +12,12 @@ import org.bukkit.entity.Player;
 public class Command_clearparticles extends TFM_Command 
 {        
     @Override
-    public boolean run(CommandSender sender, Player sender_p, Command cmd, String commandLabel, String[] args, boolean senderIsConsole) {
-        for (int i = 0; i <= 80; i++)
+    public boolean run(CommandSender sender, Player sender_p, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
+    {
         {    
-            TFM_Util.bcastMsg("");
+        server.dispatchCommand(sender, "wildcard gcmd ? pp clear");
         }
-        TFM_Util.adminAction(sender.getName(), "Clearing all particles ", true);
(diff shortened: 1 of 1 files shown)


src/me/StevenLawson/TotalFreedomMod/Commands/Command_clearparticles.java
```diff
@@ -0,0 +1,23 @@
...
+
+@CommandPermissions(level = AdminLevel.SUPER, source = SourceType.BOTH)
+@CommandParameters(description = "Particle Cleanup", usage = "/<command>", aliases = "cp")
+public class Command_clearparticles extends TFM_Command 
+{        
+    @Override
+    public boolean run(CommandSender sender, Player sender_p, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
+    {
+        {    
+        server.dispatchCommand(sender, "wildcard gcmd ? pp clear");
```
DUFCLiam: Isn't using dispatchCommand gonna get it denied?

DUFCLiam: - Avoid using org.bukkit.Server.dispatchCommand(). Commits that make use of it will likely be rejected.

LegendIsAwesomes:
I don't think you notice, the issue has to be accepted before a PR can be made.

DUFCLiam:
Well me and adam just say "YOLO BIT*ES"

DUFCLiam:
And make the PR anyways ;)

src/me/StevenLawson/TotalFreedomMod/Commands/Command_clearparticles.java
OnlineRepz: Prozza doesn't like using ``` server.dispatchCommand(); ```

Please find another way to avoid this

LegendIsAwesomes:
Well then, it'll get denied.

DUFCLiam:
Cool ;)

src/me/StevenLawson/TotalFreedomMod/Commands/Command_clearparticles.java
DUFCLiam: @OnlineRepz, I think in this case he has to use `server.dispatchCommand();`

PR #976, opened 2015-12-21
Changed Jumpad Block

kanedabeast:
I think we should change the jumpad block only because sometimes people use wool and it's really annoying for them, sometimes half of them are not even superadmins so i have changed the block to glass. Reason 1: Because hardly anyone uses it for a ground mainly for walls.


ReelPattytothewatty:
Kane, Please make an issue, if it gets accepted, then you can make a Pull Request.
Thanks

Wild1145:
Please create an associated issue, and only resolve one issue per PR. Here you are also adding a TPAS command that is both unnecessary and completely pointless as it mimics the saconfig command setup.

DUFCLiam:
@kanedabeast, Please may you remove the TPaS system please.

JeromSar:

PR #985, opened 2015-12-22
Added "Opping all players on TotalFreedom"

AlphaGaming:


DUFCLiam:
I do not think this should be added.

Wild1145:
Not necessary as TFM is a multi-server plugin. If you are going to make a change it needs an associated issue and I would advise getting the server name from the config instead so as not to cause a shit storm on other servers without a good reason.

ValenciaOrange:
Honestly, I think we should just keep it short like "Opping all players" and not "Opping all players that is playing on the server" or something like that

Adamx120:
I aggre with Wild.

JeromSar:
This PR does not meet the [pull request requirements](https://github.com/TotalFreedom/TotalFreedomMod/wiki/Contributing#pull-request-requirements) and thus is closed.

PR #1023, opened 2015-12-28
Set command output gamerule to false.

### LINKED ISSUE #1018: Disable command feedback gamerule
/playsound is a rather spam-filled command as it is widely used and can be spammed to "clear" the chat, therefor making you scroll. Normally, this isn't a huge problem when its used about one time, but when two or more people use it frequently, it can be rather annoying, especially if a staff member is also doing so.

DUFCLiam:
#1018

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/TotalFreedomMod.java
+++ b/src/me/StevenLawson/TotalFreedomMod/TotalFreedomMod.java
@@ -98,6 +98,9 @@ public class TotalFreedomMod extends JavaPlugin
         TFM_Util.createBackups(SUPERADMIN_FILENAME);
         TFM_Util.createBackups(PERMBAN_FILENAME);
 
+        //Stops command feedback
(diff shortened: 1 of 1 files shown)


JeromSar:
@DUFCLiam This is the wrong approach. As the [contributing guidelines](https://github.com/TotalFreedom/TotalFreedomMod/wiki/Contributing) (_please read them!_) dictate. The use of `dispatchCommand()` should be avoided at all times. In this case, both Bukkit and TFM have an API for handling gamerules. I suggest you use TFM's API which sets a couple of gamerules by default.

LegendIsAwesomes:
People really need to start reading the contributing guidelines, I decided every time I make a new PR I read the guidelines before opening it.

JeromSar:
@LegendIsAwesomes That's probably a good idea. :)

DUFCLiam:
@JeromSar I did I could not find a way to get around this.

LegendIsAwesomes:
@DUFCLiam Theres multiple ways around this

PR #1426, opened 2015-12-31
Update buildcreator.default.properties

ghost:


ReelPattytothewatty:
Are you trying to sneak in this, so when someone creates one, it doesnt build.
Stop SPAMMING

JeromSar:
@ReelPattytothewatty Don't feed the trolls.

PR #1463, opened 2016-01-25
Auto enable commandspy

MrImmaculate:
While still allowing it to be toggled with the command.

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/TFM_PlayerData.java
+++ b/src/me/StevenLawson/TotalFreedomMod/TFM_PlayerData.java
@@ -104,7 +104,7 @@ public class TFM_PlayerData
     private boolean allCommandsBlocked = false;
     private boolean verifiedSuperadminId = false;
     private String lastCommand = "";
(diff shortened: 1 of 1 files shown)


DUFCLiam:
Agreed it can just see if it was enabled on last login so you do not have to do /cmdspy everytime

Wild1145:
I have suggested this to @JeromSar previously however I believe he wanted to leave it how it was and have it "Opt-In" not "Opt-Out".

I do like the idea however of having it "Opt-Out" because a lot of people (Me especially) forgets to enable command spy when they login, then get problems trying to figure out what happened when... This just makes it easier.

PacksGamingHD:
I vouch for this idea

`test`

MrImmaculate:
This does not enable cmdspy, I am going to close this PR and open a new one that actually does it.

JeromSar:
@MrImmaculate Instead, why not append your changes to this PR to avoid cluttering the repository?

JeromSar:

PR #1481, opened 2016-02-07
Fixes UUIDs for /gadmin

ThePheonixBlader:
1. I set the substring from 4 to 11 because if it was 4, it would only show "dead". If it was 11, it would show "deadbeef-" then a letter and a number after the dash. For example, when I tested this, it showed this:
"deadbeef-a8" as my UUID for /gadmin list. It actually works too, I tested one of the sub commands on myself and it worked fine.

2. In /gadmin op <targethash> it would set the to non-op, so I fixed that too so now it ops the player, not deops.


LegendIsAwesomes:
Associated issue?

JeromSar:

PR #1513, opened 2016-03-01
Update TitleRank.java TFM 5.0

DUFCLiam:


DUFCLiam:
Not tested due to TFM 5.0 being incomplete and  will not build.

Issue #1494, opened 2016-02-13
Announce mob purges


OutOfYourMind:
When you do /mp, it doesn't announce mob purges in the chat, it only messages you with how many mobs were removed. I think we should make it so that it announces when the admins does /mp, and if this issue gets accepted, I will make the pull request. What I mean by announce is a  TFM_Util.adminAction(sender.getName() addition when the player does the command.

PieGuy7896:
Why would you need to broadcast this in the first place?

JeromSar:
@OutOfYourMind Good idea.

PR #1518, opened 2016-03-05
Issue #1494 Announce Mob Purges.

OutOfYourMind:
I made the PQ to announce the mob purges.


Wild1145:
While the code looks solid and the @CLAassistant has passed I dont think this follows the contributing guidelines as its only a single line change, and they dont tend to be merged due to the amount of work and complexion that goes into merging a single line commit. As this is only a single commit that doesnt look to need any changes @JeromSar may well make an exception, but I dont think it technically follows the guidelines.

DUFCLiam:
I agree with you @Wild1145 this should be denied.

OutOfYourMind:
Alright.

Issue #1524, opened 2016-03-07
(TFM 5.0) Cannot block players commands


LegendIsAwesomes:
It says they're a super admin, I might be able to fix this.

Wild1145:
Can you explain how you re-created this? I cannot re-create it in game...

LegendIsAwesomes:
A normal op was spamming command, I did /blockcmd (his name) and it came up saying he is a admin when he wasn't.

@Wild1145

Wild1145:
I just tested it on Farrah_, I removed his admin, reloaded the adminlist, checked he was not admin and blocked his commands. TFM blocked the commands and proceeded to stop him being able to run commands... I am not sure why you had the bug (that should be a bug by the looks of #1525) and yet worked for me...

JeromSar:
It was fixed some time ago.

PR #1525, opened 2016-03-07
Correct if statement

Adaptified:
Solves #1524, check if the target player is an admin instead of the sender.


Wild1145:
Strange... I was unable to re-create #1524 and yet this should actually fix said issue...

LegendIsAwesomes:
Goddamnit I was gonna make this. @Adaptified Your fast xD.

JeromSar:
This was already fixed. Thanks for your contribution. :)

PR #1603, opened 2016-04-04
Resolves #752 - Create command "/wipewarpdata"

### LINKED ISSUE #752: Wipe Warps
Sometimes the server gets overload with useless warps such as /warp ssdasdwds 
When the map gets wiped, warps are useless because they lead to nowhere. 
/wipewarpdata could be useful to delete warps

http://totalfreedom.boards.net/thread/21578/warp-names-spam-warps

ghost:
Created command "/wipewarpdata", in method, delete folder warps in Essentials folder

### DIFF
--- a/src/me/StevenLawson/TotalFreedomMod/Command_wipewarpdata.java
+++ b/src/me/StevenLawson/TotalFreedomMod/Command_wipewarpdata.java
@@ -1,0 +1,29 @@
+package me.StevenLawson.TotalFreedomMod.Commands;
+
+import java.io.File;
+import me.StevenLawson.TotalFreedomMod.TFM_Util;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+
+@CommandPermissions(level = AdminLevel.SENIOR, source = SourceType.ONLY_CONSOLE)
+@CommandParameters(description = "Deletes all warps from the server", usage = "/<command>")
+public class Command_wipewarpdata extends TFM_Command
+{
+    @Override
+    public boolean run(CommandSender sender, Player sender_p, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
+    {
+        if (!server.getPluginManager().isPluginEnabled("Essentials"))
+        {
+            playerMsg("Essentials is not enabled on this server");
+            return true;
+        }
+
+        TFM_Util.adminAction(sender.getName(), "Wiping all warps from the server", true);
+
+        TFM_Util.deleteFolder(new File(server.getPluginManager().getPlugin("Essentials").getDataFolder(), "warps"));
+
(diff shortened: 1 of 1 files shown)


DUFCLiam:
@Adamx120, Do not delete the file just clear it.

RobinGall2910:
Deleting the file would just regenerate, and can you _please_ stop tagging on every single post you type on this GitHub repository? @DUFCLiam

Wild1145:
You if you delete the file while the server is running you do have the scope for issues, you would need to disable Essentials, delete the file and re-enable essentials. This would however require Essentials to **HAVE** to go through its initial start up and not just be "unblocked" like TFM plugin disabling appears to do (With WorldEdit at least it does, I Found that today)

Adamx120:
@Wild1145 But if a player creates a warp, the warp file regenerates

DUFCLiam:
@Adamx120, May you make it clears the file. Add something like `.clear` after locating the file? Rather than deleting it?

DUFCLiam:
Also make it 5.0.

OxLemonxO:
@DUFCLiam It would just regenerate after someone does /setwarp <warp>

src/me/StevenLawson/TotalFreedomMod/Command_wipewarpdata.java
```diff
@@ -0,0 +1,29 @@
...
+        {
+            playerMsg("Essentials is not enabled on this server");
+            return true;
+        }
+
+        TFM_Util.adminAction(sender.getName(), "Wiping all warps from the server", true);
+
+        TFM_Util.deleteFolder(new File(server.getPluginManager().getPlugin("Essentials").getDataFolder(), "warps"));
+
+        playerMsg("Deleting warps success.");
```
Wild1145, line 26: This does not check if the warps actually are purged. Without it this would always be displayed even if there is an error, some sort of check is needed really.

src/me/StevenLawson/TotalFreedomMod/Command_wipewarpdata.java
```diff
@@ -0,0 +1,29 @@
+package me.StevenLawson.TotalFreedomMod.Commands;
+
+import java.io.File;
+import me.StevenLawson.TotalFreedomMod.TFM_Util;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+
+@CommandPermissions(level = AdminLevel.SENIOR, source = SourceType.ONLY_CONSOLE)
```
Wild1145, line 9: Any reason it has to be console only? If its senior only already maybe we could allow it in game? Just an idea, something for further discussion.

Wild1145 (changes requested):
One change to be made, one thing to be reviewed and then testing will be needed.

PR #1606, opened 2016-04-06
Add onlinemode on/off notification - closes #778/#779

### LINKED ISSUE #778: MotD Online/Offline mode Notif

RobinGall2910:
Since the CLA wouldn't be signed and that TFM 5.0 is out, well I fixed it if you still want it in.

I wouldn't make it configurable, sorry. Unless if I had some time..

Duplicate of #779, just that the CLA is signed and the code is inserted into the appropriate file.


RobinGall2910:
#778 has been closed, didn't notice.

PR #1772, opened 2016-06-15
Update Rank.java

TomCreeper:
Could I please download the mod?


CreeperSeth:
Question: Why would you change the color of senior admin and then say  
> Could I please download the mod?

PR #1775, opened 2016-06-23
A small file error needs to be accepted!

Superior-tL:
Fixed the command needs to be accepted because the command makes a lot of errors, needs to be added, @JeromSar  @Wild1145


AvalancheYT:
It doesn't affect other commands.

LegendIsAwesomes:
Please read the guidelines, as one word changes or small changes can not be accepted, this PR also doing have an issue with the label 'Accepted' therefore again, it can not be accepted.

Contributing Guidelines: https://github.com/TotalFreedom/TotalFreedomMod/wiki/Contributing

Wild1145:
We couldn't actually have aliases anyway if I recall, we were trying to remove them all anyway. But what he said above.

JeromSar:
This is not the way TFM works with aliases.

Issue #1791, opened 2016-07-21
Block /summon


CrafterSmith12:
From thread: http://totalfreedom.boards.net/thread/42772/structure-blocks

Wild1145:
This should be done in a config change server side. I dont understand why this is being raised as an issue here...

notanoperator:
https://github.com/TotalFreedom/TotalFreedomMod/pull/1792/commits/75b8e6ab2379e005ce60a471fbd812db4d70d665

robingall2910:
This should be just in the server, and the server only, not the plugin.

Wild1145:
Resolved as part of #2002

PR #1792, opened 2016-07-21
block summon

notanoperator:


Wild1145:
What issue does this resolve?

notanoperator:
In reply to Wild - https://github.com/TotalFreedom/TotalFreedomMod/issues/1791

LegendIsAwesomes:
Please read the contributing guidelines: https://github.com/TotalFreedom/TotalFreedomMod/wiki/Contributing , as this is just a small config edit and therefore is against the contributing guidelines as it clearly states, in the words of Jerom (Prozza), thank you for contributing.

src/main/resources/config.yml
```diff
@@ -120,6 +120,7 @@ blocked_commands:
   - 'n:b:/setjail:_'
   - 'n:b:/tfbanlist:Use /banlist to see and wipe the ban list'
   - 'n:b:/tfipbanlist:Use /banlist to see and wipe the ban list'
+  - 'n:b:/summon:This command has been blocked for security reasons'
```
JeromSar: I think this should just be an empty '_' message. Since most of these commands are blocked for "security measures". The empty message implies this.

Wild1145 (changes requested):
Please make the changes addressed earlier before this PR can be merged.

PR #1794, opened 2016-07-21
Add /uall and /dtoggle

### LINKED ISSUE #1793: Add /uall and /dtoggle
Mark recently added Libsdisguises to the server and we need /uall and /dtoggle I am re-adding those commands now

script-head:
mark added libs disguise on the server and we needed /uall and /dtoggle back so I re-added them and tested the commands and they worked perfectly This resolves #1793


Wild1145:
You should really also add a soft depend on the plugin, else this command will not work. You also then need to check if the plugin is enabled and working so as not to cause errors...

CreeperSeth:
Did what Wild recommended that I do

CreeperSeth:
Closed it as Lemon put the iceing on the cake see #1800

PR #1799, opened 2016-07-27
Add a Master Builder Rank

Inergivi:


Inergivi:
#7198

jjkatz:
I believe u can only make a PR if u have the issue accepted

robingall2910:
PR is going to be denied since it's less than 10 lines of code, and the issue isn't approved.

Wild1145:
This would also cause some major permissions issues...

PR #1803, opened 2016-08-01
Telnet clan. rank

fionn123:
I believe there should be a telnet clan rank in game like a saconfig setrank (username) telnet_clan_admin or something. As I believe, this would stop a lot of confusion and make it easier, for admin chat, I'd suggest [TCA] and for tag I'd suggest [Telnet-ClanForge] or something else, up to prozza, but what do you think?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,9 +60,9 @@
         <dependency>
             <groupId>org.spigotmc</groupId>
             <artifactId>spigot</artifactId>
-            <version>1.9-R0.1-SNAPSHOT</version>
+            <version>1.10</version>
             <scope>system</scope>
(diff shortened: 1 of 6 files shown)


robingall2910:
No, since this is a TF exclusive rank only, it wouldn't happen unless you made this certain rank only for TFM mod on TF and would've not commited/coded the certain rank on this github repository.

btw why the hell did you add commits

Wild1145:
As there is no change in permissions there is no need.

Also, this mod is designed for TF so any additions that are made should focus on TF and no other servers.

Wild1145:
Also, this PR is complete crap... Like complete and utter crap...

JeromSar:
@fionn123 I don't believe you understand how PRs work.

PR #1894, opened 2016-10-30
W

### LINKED ISSUE #1892: /wildcard STILL NOT WORKING
Title is pretty self explanatory, according to OxLemonxO the version of the command on the server works but the code in github is still broken.

WarrioR151:
F


JeromSar:
@WarrioR151 Stop doing what you're doing and learn how to use Github _before_ doing anything else that doesn't make sense, please.

PR #1908, opened 2016-11-12
Fix #1872 (Command_dtoggle.java message mismatch)

### LINKED ISSUE #1872: Command_disguisetoggle.java
Line 37 of the file located at: TotalFreedomMod/src/main/java/me/totalfreedom/totalfreedommod/command/Command_disguisetoggle.java from the branch tfm5.1-mc1.10 reads this:

`msg("Enabled " + (DisallowedDisguises.disabled ? "enabled." : "disabled."));`

This results in an error.
When the plugin is disabled/enabled it displays the opposite message of it's state,  that is, when the plugin is enabled it says it messages you that it is disabled, and vice versa. Also, when you are messaged it does not message you "Disguises enabled" as it should, instead it messages you "Enabled enabled" and "Enabled disabled", which in my opinion doesn't make much sense. 

To fix these errors simply changed line 37 to read this:

`msg("Disguises " + (DisallowedDisguises.disabled ? "disabled." : "enabled."));`

or

`msg("Disguises " + (!DisallowedDisguises.disabled ? "enabled." : "disabled."));`

Thanks for reading!

ghost:
Fixed message mismatch.

### DIFF
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_disguisetoggle.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_disguisetoggle.java
@@ -1,6 +1,5 @@
 package me.totalfreedom.totalfreedommod.command;
 
-import me.libraryaddict.disguise.DisallowedDisguises;
 import me.totalfreedom.totalfreedommod.rank.Rank;
 import me.totalfreedom.totalfreedommod.util.FUtil;
 import org.bukkit.command.Command;
@@ -22,7 +21,7 @@ public class Command_disguisetoggle extends FreedomCommand
             return true;
         }
 
-        FUtil.adminAction(sender.getName(), (DisallowedDisguises.disabled ? "Enabling" : "Disabling") + " Disguises", false);
+        FUtil.adminAction(sender.getName(), (!plugin.ldb.isDisguisesEnabled() ? "Enabling" : "Disabling") + " Disguises", false);
 
         if (plugin.ldb.isDisguisesEnabled())
         {
@@ -34,7 +33,7 @@ public class Command_disguisetoggle extends FreedomCommand
             plugin.ldb.setDisguisesEnabled(true);
         }
 
-        msg("Disguises " + (!DisallowedDisguises.disabled ? "enabled." : "disabled."));
+        msg("Disguises " + (plugin.ldb.isDisguisesEnabled() ? "enabled." : "disabled."));
 
         return true;
     }
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_disguisetoggle.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_disguisetoggle.java
@@ -34,7 +34,7 @@ public class Command_disguisetoggle extends FreedomCommand
             plugin.ldb.setDisguisesEnabled(true);
         }
 
-        msg("Enabled " + (DisallowedDisguises.disabled ? "enabled." : "disabled."));
+        msg("Disguises " + (!DisallowedDisguises.disabled ? "enabled." : "disabled."));
 
         return true;
     }
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_gcmd.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_gcmd.java
@@ -1,56 +1,83 @@
 package me.totalfreedom.totalfreedommod.command;
 
+import java.util.Arrays;
+import java.util.List;
 import me.totalfreedom.totalfreedommod.rank.Rank;
 import org.apache.commons.lang3.StringUtils;
+import org.bukkit.ChatColor;
 import org.bukkit.command.Command;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
 
-@CommandPermissions(level = Rank.SUPER_ADMIN, source = SourceType.BOTH, blockHostConsole = true)
+@CommandPermissions(level = Rank._ADMIN, source = SourceType.ONLY_IN_GAME, blockHostConsole = true)
 @CommandParameters(description = "Send a command as someone else.", usage = "/<command> <fromname> <outcommand>")
 public class Command_gcmd extends FreedomCommand
 {
 
+    public static final List<String> BLOCKED_COMMANDS = Arrays.asList(
+            "wildcard",
+            "gtfo",
+            "doom",
+            "saconfig",
+            "gcmd",
+            "glist",
+            "tban",
+            "tempban"
+    );
+
     @Override
     public boolean run(CommandSender sender, Player playerSender, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
     {
-        if (args.length < 2)
-        {
-            return false;
-        }
-
-        final Player player = getPlayer(args[0]);
-
-        if (player == null)
-        {
-            sender.sendMessage(FreedomCommand.PLAYER_NOT_FOUND);
-            return true;
-        }
-
-        final String outCommand = StringUtils.join(args, " ", 1, args.length);
-
-        if (plugin.cb.isCommandBlocked(outCommand, sender))
-        {
-            return true;
-        }
-
         try
         {
-            msg("Sending command as " + player.getName() + ": " + outCommand);
-            if (server.dispatchCommand(player, outCommand))
-            {
-                msg("Command sent.");
-            }
-            else
-            {
-                msg("Unknown error sending command.");
-            }
-        }
-        catch (Throwable ex)
-        {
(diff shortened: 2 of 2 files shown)


Wild1145:

Wild1145 (commented):
Testing required.

Commodore64x:
You are kidding me.... "Testing required" This is not even a major code change!! I literally just added a ! then changed Enabled to Disguises.

JeromSar:
@Commodore64x You might think that for your personal projects, but on this repository, we like to uphold a high code quality. Therefore, testing is mandatory.

Commodore64x:
@JeromSar but I mean its a 1 line code change that is guaranteed to work...

Wild1145:
@Commodore64x The point is, it is not guaranteed to work...

Commodore64x:
@Wild1145, at this point you should be able to read code, and at a simplistic level like this where the code is easy to read you can tell it will work.

mayokek:
@Commodore64x, sometimes, some simple codes doesn't work.

Wild1145:
@Commodore64x And I can not guarantee 100% that said code would work. Thats the point here.

JeromSar:
@Commodore64x No more discussion now. We have good workflow that we adhere to, and we will continue to use that.

lemonsked:
Can you change `!DisallowedDisguises.disabled` to the `isDisguisesEnabled` method in LibsDisguises bridge?

Wild1145 (dismissed):
Please make the requested changes that were mentioned earlier.

ghost:
Ok, I guess code simplification is for the best.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_disguisetoggle.java
```diff
@@ -22,7 +21,7 @@ public boolean run(CommandSender sender, Player playerSender, Command cmd, Strin
             return true;
         }
 
-        FUtil.adminAction(sender.getName(), (DisallowedDisguises.disabled ? "Enabling" : "Disabling") + " Disguises", false);
+        FUtil.adminAction(sender.getName(), (!plugin.ldb.isDisguisesEnabled() ? "Enabling" : "Disabling") + " Disguises", false);
```
lemonsked, line 24: You do not need to invert this.

ghost:
@OxLemonxO no, you are wrong. The code works find how it is, if I removed the ! then the messages would be mismatched again... The code line would detect if the plugin was already enabled and then print a message saying it was enabled again... It is meant to detect whether the plugin is disabled or enabled then print the correct message. If the plugin was enabled which evaluates to 'true' then it would say {player} - enabling disguises, which would conflict with the message you get given in chat.

mayokek:
Actually @Commodore64x, Lemon is correct.

ghost:
@AvalancheYT you probably haven't even looked at the code, much less tested it. If you had then you would know that it works, removing the ! will get us back where we started with the messages being stuffed up!

src/main/java/me/totalfreedom/totalfreedommod/command/Command_disguisetoggle.java
ghost: Yeah you kinda do...

ghost:
Actually to further prove this: the original method was DisallowedDisguises.disabled then it was changed to !plugin.ldb.isDisguisesEnabled(). That method reads as follows: 
`public boolean isDisguisesEnabled()
    {
        return !DisallowedDisguises.disabled;
    }`
Notice how it returns the opposite value of the original method? Therefore, in order to work correctly there needs to be a ! to cancel the ! in the IsDisguisesEnabled Method. Proof enough?

mayokek:
@Commodore64x, ! doesn't cancel !
```java
public boolean isDisguisesEnabled()
{
    return !DisallowedDisguises.disabled;
}
```
will come out as enabled.

ghost:
@AvalancheYT stop contradicting me. I know for a FACT that what Lemon suggested is wrong (I tested it). I suppose you haven't thought about the ! either. Let me explain, lets say boolean gupta = true;
so if you did if (gupta) { code} it would execute the code. If you did if (!gupta) { code} gupta now equates to false and the code will no longer execute.
If you have a method, say 
```
public void gupta() {
return !gupta;
}

```
but you want to get the actual value of gupta, not with ! in front of it, you invert the method again so just by doing !gupta() it equates to  
```
public void gupta() {
return gupta;
}
```

mayokek:
@Commodore64x, I don't believe you have tested this.
```java
if (plugin.ldb.isDisguisesEnabled())
{
    //code
}
```
It would come out as enabled (true) and if you have -
```java
if (!plugin.ldb.isDisguisesEnabled())
{
    //code
}
```
then, it would come out as disabled (false)

Commodore64x:
@AvalancheYT I have tried to be calm, but now I have lost my patience. Please stop commenting on this PR. You obviously have little experience with java, and do not know what you are talking about.

ghost:
What you say is complete bs, but regardless let me help you understand. The code reads: 
```
if (!plugin.ldb.isPluginEnabled())
        {
            msg(ChatColor.RED + "LibsDisguises is not enabled.");
            return true;
        }

        FUtil.adminAction(sender.getName(), (!plugin.ldb.isDisguisesEnabled() ? "Enabling" : "Disabling") + " Disguises", false);

        if (plugin.ldb.isDisguisesEnabled())
        {
            plugin.ldb.undisguiseAll(true);
            plugin.ldb.setDisguisesEnabled(false);
        }
        else
        {
            plugin.ldb.setDisguisesEnabled(true);
        }

        msg("Disguises " + (plugin.ldb.isDisguisesEnabled() ? "enabled." : "disabled."));

        return true;
    }
}
```

Apart from the behind the scenes stuff, there are two messages that display themselves in chat: One that is messaged DIRECTLY to you, and one that is announced in chat. The one that is sent directly to you tell you what happened, e.g. I just enabled disguises and it will messaged me Disguises enabled. What I believe you are having trouble with is ternary operators. if the given statement is true then the first given option is executed, and if the statement is false then it executes the second. If disguises have been enabled the method will detect that it is enabled and msg the player that it has been enabled. Now for the message that is announced in chat, FUtil.adminAction(sender.getName(), (!plugin.ldb.isDisguisesEnabled() ? "Enabling" : "Disabling") + " Disguises", false);
This ! reverses the output of the method, e.g. gupta = true but !gupta will equal false. So, if we did the method lemon suggested the code would check if the plugin was enabled. Lets consider a case where Disguises are disabled, the code checks if they are enabled, which it is not, so as the method has returned a false value, the second option is executed, resulting in the messaged Disguises Disabling, which, since Disguises are already disabled is a bit weird. The same would happen if disguises were already enabled. Enter the ! in the current code. It makes the method return the OPPOSITE value from what it returns, for example if the plugin was enabled, the code checks if the plugin is enabled, sees the ! and turns its true value to false, which executes the second option.
This is a fairly lengthy explanation but I hope you understand now.

ghost:
No more things to be changed @Wild1145

mayokek:
@Commodore64x, are you sure that you are correct?
```java
if (plugin.ldb.isDisguisesEnabled()) // Checks if the plugin/disguises is enabled (which is enabled)
{
    plugin.ldb.undisguiseAll(true); // Undisguises everyone before disabling
    plugin.ldb.setDisguisesEnabled(false); // Disable the plugin/disguises
}
else // The plugin is disabled.
{
    plugin.ldb.setDisguisesEnabled(true); // Enable the plugin/disguises
}
```

ghost:
That was given code, I am only fixing the issue that is presented.

ghost:
However since the purpose of the code is to enable/disable the plugin it does seem strange, my idea was that it is for when the TFM cannot resolves LibsDisguises.

AvalancheYT:
@Commodore64x, the given code is an example of how it worked.
So it should be like
```java
FUtil.adminAction(sender.getName(), (plugin.ldb.isDisguisesEnabled() ? "Enabling" : "Disabling") + " Disguises", false);
```

mayokek:
Oh, now I just realised. You are correct.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_disguisetoggle.java
mayokek: @OxLemonxO, the code is correct. If the disguises are disabled, it would need to say enabling not disabling.

ghost:
@Wild1145 please remove the Needs changes tag, unless anyone request any more.

lemonsked:
Oh sorry, my mistake.

Wild1145 (commented):
Testing is required.

Wild1145:
The original repo has been destroyed and merge conflicts exist.

PR #1940, opened 2016-12-20
Update BlockBlocker.java

fIuffy:
Anti-Crash Chests...


mayokek:
Edit: Seems that I misread the code.
What issue does this solve?

fIuffy:
Oh wait never mind, idk why I even sent this in the first place -,-

PR #1956, opened 2017-01-12
Create Command_clearchat.java

ghost:
**What is this and why I created this**
This command is as it says in its own name. It clears the public chat. This is good for serial advertisers who like to take advantage of the server to join just to advertise; spam ads, etc.

I have made it so only online players in-game receive blank chat messages (to clear the chat) to prevent spam in console and telnet (as they have to be a 'player').

To prevent abuse and misuse of the command, only senior admins will be able to use this command.

### DIFF
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_clearchat.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_clearchat.java
@@ -3,9 +3,7 @@ package me.totalfreedom.totalfreedommod.command;
 import me.totalfreedom.totalfreedommod.rank.Rank;
 import me.totalfreedom.totalfreedommod.util.FUtil;
 
-import org.bukkit.Bukkit;
-import org.bukkit.ChatColor;
-import org.bukkit.GameMode;
+import org.bukkit.*;
 import org.bukkit.command.Command;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_clearchat.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_clearchat.java
@@ -1,0 +1,35 @@
+package me.totalfreedom.totalfreedommod.command;
+
+import me.totalfreedom.totalfreedommod.rank.Rank;
+import me.totalfreedom.totalfreedommod.util.FUtil;
+
+import org.bukkit.Bukkit;
+import org.bukkit.ChatColor;
+import org.bukkit.GameMode;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+
+@CommandPermissions(level = Rank.SENIOR_ADMIN, source = SourceType.BOTH, blockHostConsole = true, aliases = "cc")
+@CommandParameters(description = "Clear the public chat.", usage = "/<command>")
+public class Command_clearchat extends FreedomCommand
+{
+    @Override
+    public boolean run(CommandSender sender, Player playerSender, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
+    {
+        Player p = (Player) sender;
+        
+        // Checks and sends ONLY a blank message to all players on the server (to prevent console spam)
+        for (Player player : server.getOnlinePlayers())
+        {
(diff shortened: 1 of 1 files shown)


mayokek:
And this resolves to what issue?

Wild1145:

src/main/java/me/totalfreedom/totalfreedommod/command/Command_clearchat.java
```diff
@@ -0,0 +1,33 @@
+package me.totalfreedom.totalfreedommod.command;
+
+import me.totalfreedom.totalfreedommod.rank.Rank;
+import me.totalfreedom.totalfreedommod.util.FUtil;
+
+import org.bukkit.*;
```
Wild1145, line 6: Please do not use wildcard imports.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_clearchat.java
```diff
@@ -0,0 +1,33 @@
...
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+
+@CommandPermissions(level = Rank.SENIOR_ADMIN, source = SourceType.BOTH, blockHostConsole = true, aliases = "cc")
+@CommandParameters(description = "Clear the public chat.", usage = "/<command>")
```
Wild1145, line 12: This would not be quite accurate, as it would clear all chats, both public and admin chat.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_clearchat.java
```diff
@@ -0,0 +1,33 @@
...
+    {
+        Player p = (Player) sender;
+        
+        // Checks and sends ONLY a blank message to all players on the server (to prevent console spam)
+        for (Player player : server.getOnlinePlayers())
+        {
+            // Run code inside the loop 100 times
+            for (int i = 0; i < 100; i++)
+            {
+                player.sendMessage("");
```
Wild1145, line 26: Is there no cleaner way to do this? This will cause massive spam in log files both server and client side...?

src/main/java/me/totalfreedom/totalfreedommod/command/Command_clearchat.java
```diff
@@ -0,0 +1,33 @@
...
+            }
+        }
+        
+        Bukkit.broadcastMessage(ChatColor.BLUE + "Chat cleared by " + p.getName() + ".");
```
Wild1145, line 30: This should use the TFM Admin action method.

Wild1145 (changes requested):
Please make the changes as commented.

robingall2910:
This shouldn't be pushed. From TF's point of view, you can just press ``D`` while holding ``F3`` and that would simply clear up the chat and your command history.

More info: http://totalfreedom.boards.net/thread/48702/clear-chat

ghost:
It is possible to make a command like this without spamming logs but most people object to the idea.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_clearchat.java
Hockeyfan360: @Wild1145 The server does not log when plugins send players messages. This will spam client side logs, but not server side logs. Admins on telnet will not be affected. Most people don't check client logs anyways, to be honest, so there isn't much to worry about.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_clearchat.java
```diff
@@ -0,0 +1,33 @@
+package me.totalfreedom.totalfreedommod.command;
+
+import me.totalfreedom.totalfreedommod.rank.Rank;
+import me.totalfreedom.totalfreedommod.util.FUtil;
+
+import org.bukkit.*;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
```
robingall2910: these imports aren't needed if you're using the wildcard btw, so just remove the wildcard instead

ghost:
@Wild1145 Multiple admin votes have decided that this feature shouldn't be implemented and there are conflicts.

Wild1145:

PR #1957, opened 2017-01-14
Create Command_advertise.java

ghost:
An advertising command with a 10-minute cooldown.

**Suggested at:**
http://totalfreedom.boards.net/thread/25664/advertisement-command

**The thread:**
Usage: /advertise <advert>
Before you start saying 
'Object ops will spam it'
Just like the rules say, once every 10 minutes. Once the op uses the command they need to wait 10 minutes before they can post another one.
It could look like 
[Advertisement]"Player"]join play.totalfreedom.me
If they attempt to spam it will say "You need to wait another 5:40 before you can post another advertisement.
"I object ops will abuse it by putting silly messages" Erm. We're admins, if they do we'll do the correct punishment.

### DIFF
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
@@ -26,7 +26,7 @@ public class Command_advertise extends FreedomCommand
         
         if (cooldown.contains(playerSender))
         {
-            playerSender.sendMessage(ChatColor.RED + "You must wait 10 minutes to make another advertisement.");
+            msg("You must wait 10 minutes to make another advertisement.");
             return true;
         }
         
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
@@ -30,7 +30,7 @@ public class Command_advertise extends FreedomCommand
             return true;
         }
         
-        if (args.length > 1)
+        if (args.length == 0)
         {
             return false;
         }
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
@@ -23,11 +23,10 @@ public class Command_advertise extends FreedomCommand
     @Override
     public boolean run(CommandSender sender, Player playerSender, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
     {
-        Player p = (Player) sender;
         
-        if (cooldown.contains(p))
+        if (cooldown.contains(playerSender))
         {
-            p.sendMessage(ChatColor.RED + "You must wait 10 minutes to make another advertisement.");
+            playerSender.sendMessage(ChatColor.RED + "You must wait 10 minutes to make another advertisement.");
             return true;
         }
         
@@ -40,15 +39,15 @@ public class Command_advertise extends FreedomCommand
         Bukkit.broadcastMessage(ChatColor.GREEN + "[Advertisement] " + ChatColor.GOLD + sender.getName() + ChatColor.GREEN + "] " + ChatColor.DARK_GREEN + message);
         
         // Begin cooldown
-        cooldown.add(p);
+        cooldown.add(playerSender);
         
         Bukkit.getServer().getScheduler().scheduleSyncDelayedTask(plugin, new Runnable()
         {
             public void run()
             {
                 // End cooldown
-                p.sendMessage(ChatColor.GRAY + "You may now make another advertisement.");
-                cooldown.remove(p);
+                msg("You may now make another advertisement.");
+                cooldown.remove(playerSender);
             }
         }, 600L * 20L); 
     }
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
@@ -30,13 +30,13 @@ public class Command_advertise extends FreedomCommand
             return true;
         }
         
-        if (args.length < 1)
+        if (args.length > 1)
         {
             return false;
         }
         
         String message = StringUtils.join(args, " ", 0, args.length);
-        Bukkit.broadcastMessage(ChatColor.GREEN + "[Advertisement] " + ChatColor.GOLD + sender.getName() + ChatColor.GREEN + "] " + ChatColor.DARK_GREEN + message);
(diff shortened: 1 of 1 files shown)


src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
```diff
@@ -0,0 +1,55 @@
...
+@CommandParameters(description = "Advertise a message every 10 minutes", usage = "/<command> <message>")
+public class Command_advertise extends FreedomCommand
+{
+    public static ArrayList<Player> cooldown = new ArrayList<Player>();
+    
+    @Override
+    public boolean run(CommandSender sender, Player playerSender, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
+    {
+        Player p = (Player) sender;
+
```
lemonsked: Just use playerSender

src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
```diff
@@ -0,0 +1,55 @@
...
+    
+    @Override
+    public boolean run(CommandSender sender, Player playerSender, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
+    {
+        Player p = (Player) sender;
+        
+        if (cooldown.contains(p))
+        {
+            p.sendMessage(ChatColor.RED + "You must wait 10 minutes to make another advertisement.");
+            return true;
```
lemonsked: use msg(message)

lemonsked (changes requested):

Wild1145:

src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
```diff
@@ -0,0 +1,54 @@
...
+        }
+        
+        if (args.length < 1)
+        {
+            return false;
+        }
+        
+        String message = StringUtils.join(args, " ", 0, args.length);
+        Bukkit.broadcastMessage(ChatColor.GREEN + "[Advertisement] " + ChatColor.GOLD + sender.getName() + ChatColor.GREEN + "] " + ChatColor.DARK_GREEN + message);
```
Wild1145: The string here would make no sense... It would be [Advertisement] wild1145 ] join my server its cool.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
```diff
@@ -0,0 +1,54 @@
...
+    public boolean run(CommandSender sender, Player playerSender, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
+    {
+        
+        if (cooldown.contains(playerSender))
+        {
+            msg("You must wait 10 minutes to make another advertisement.");
+            return true;
+        }
+        
+        if (args.length < 1)
```
Wild1145: Might want to check this. It looks like it will never work.

Wild1145 (changes requested):
Please make required changes and add an issue number.

ghost:
Issue #1958 created.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
```diff
@@ -23,11 +23,10 @@
     @Override
     public boolean run(CommandSender sender, Player playerSender, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
     {
-        Player p = (Player) sender;
         
-        if (cooldown.contains(p))
+        if (cooldown.contains(playerSender))
         {
-            p.sendMessage(ChatColor.RED + "You must wait 10 minutes to make another advertisement.");
+            playerSender.sendMessage(ChatColor.RED + "You must wait 10 minutes to make another advertisement.");
```
lemonsked: Still using player.sendMessage.
Use msg(msg)

src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
```diff
@@ -30,7 +30,7 @@ public boolean run(CommandSender sender, Player playerSender, Command cmd, Strin
             return true;
         }
         
-        if (args.length > 1)
+        if (args.length == 0)
         {
```
OxLemonxO: use = instead of == to compare integers

src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
```diff
@@ -0,0 +1,54 @@
...
+            return true;
+        }
+        
+        if (args.length == 0)
+        {
+            return false;
+        }
+        
+        String message = StringUtils.join(args, " ", 0, args.length);
+        Bukkit.broadcastMessage(ChatColor.GREEN + "[Advertisement] " + ChatColor.GOLD + sender.getName() + ChatColor.GREEN + "> " + ChatColor.DARK_GREEN + message);
```
Wild1145, line 39: Again, I dont think this is the type of formatting that will look good.

Wild1145 (changes requested):

Wild1145:
This PR will remain on hold until the associated issue is reviewed, discussed and approved.

ghost:
Ops would just use chat or /me which isn't being throttled.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
```diff
@@ -0,0 +1,54 @@
...
+{
+    public static ArrayList<Player> cooldown = new ArrayList<Player>();
+    
+    @Override
+    public boolean run(CommandSender sender, Player playerSender, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
+    {
+        
+        if (cooldown.contains(playerSender))
+        {
+            msg("You must wait 10 minutes to make another advertisement.");
```
mayokek, line 29: I reckon it should say how long left until they can execute the command.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_advertise.java
mayokek: To be honest, I believe it should be added to configuration for owner to edit it to their liking.

ghost:
Once again @Wild1145, unwanted command that ops won't use.

Wild1145:
I do however think that this could be a good idea, to only allow IP's and similar through this type of command would certainly reduce a lot of the constant monitoring admins need to do. If the above changes can be addressed, I think this could be a good addition.

JeromSar:
As I recall, we discussed the option of having /advertise before, in order to track and reduce advertising on TF.

StevenLawson (changes requested):
Rather than using a List it would be better to have a Map that stored Player -> UNIX Timestamp. Instead of using a separate thread, just have the command check the (current time - saved time) offset to make sure it is > than the minimum, and don't bother sending a message when the command can be used again.

Wild1145:

PR #1970, opened 2017-02-16
[Command] '/ask' & '/respond'

GreenProgram:
This command will have a toggle.
This can be used for OPs & Admins to test.
An OP would /ask a question and that message will
be sent to an online administrator. When an administrator views
that question, he could respond with a /respond <Target> &
it will respond to that certain OP. 
The /ask command will contain a 10 second cooldown so if a OP
tries to abuse it or he leaves /ask toggled on, it will tell him to wait 10 seconds
or un-toggle.

If this pull request gets accepted, I don't mind coding the command.


ItzLevvie:
Close? @GreenProgram

lemonsked:
I think you meant to um, make an issue, not a PR?

Issue #1830, opened 2016-08-19
Command Spy Enhancements 


Wild1145:
_There are quite a couple of issues open with regards to command spy, and I hope to more or less bring them all together into this single issue, just to keep things easy to track and to try and cut down the similar issues open._ 

- [ ] One of the additions that has been requested is to be able to activate command spy on a player by player basis, meaning you can only spy on certain players commands at a time. The ability to add & remove players from this has been suggested to reduce spam on command spy and make it easier to work with. (Original Issue #1761 )

- [ ] As part of this issue the other suggestion was to have the ability to tag certain players chats with different colours, again making Command spy easier to navigate and work with and the ability to update and amend these rules as the admins need on the server. (Original Suggestion by @DUFCLiam on issue #1761 )

- [ ] The other suggestion around command spy comes about from server reloads, currently command spy is not persistent. With the ability to add all these rules and other snazzy things it would be more useful if command spy was persistent and your settings saved somehow. Otherwise every re-load will require you to re-setup all these rules. While it will be a bit of a pain, it will save a lot of time. (Original Issue Taken from #1478 )

- [ ] The final suggestion for Command spy comes around WorldEdit, and while it is more difficult to add having the worldedit commands highlighted (Along with a preset list of other potentially "harmful" commands potentially) to flag to admins and raise their attention along side the ability to see the area of the worldedit command as part of the command. This will require a bit more thinking through but again will likely pose to be useful for server administration. (Original Issue #240 )

Wild1145:
**Updated to reflect other similar issues**

marcocorriero:
@Wild1145 if you edit worldedit we cui function, admins with we grid mod they can see the specified player selection just typing 

/showcui player

make it only admins.

it shows the Player's selection with worldedit CUI mod.

DUFCLiam:
@uyscutix:
`
This is a similar to wild's suggestion on commandspy. This is the suggestion:

When CommandSpy automatically detects a world edit command (or a command beginning with //), it would also show by the side in another colour how many blocks a player attempted to use.

Example of how it would be shown:
[WorldEdit] [284624 blocks] Griefer12345: //set dirt

This is useful, instead of having to open up the logviewer every time or so, you can see how large a world edit attempt has been by another player with the in-game command spy.

To reduce the flood/spam, if it was an obvious grief/crash attempt like "[8565748375843 blocks]" the maximum it would go is 1m and displayed like this:

[...]

PR #1980, opened 2017-03-06
Commodore64x #1830 /cmdspy

ghost:
Code for #1980


Wild1145 (changes requested):
I would probably re-structure this to first check if command spy is enabled, and then what the rank is, its just more structured instead of having if statement after if statement.

Wild1145:
Please also update the description to reflect the new issue number.

ghost:
@Wild1145 please elaborate on what you want me to change.

Wild1145:
@Commodore64x You should be checking if the user has command spy enabled first, then doing additional checks after. Also you should have a reference to the new issue (the one your issue was superseded by) in the opening message of this PR.

Wild1145:
Due to the original repo having been destroyed, I am closing this PR down as a result.

PR #2002, opened 2017-04-21
Configuration File Sync

Wild1145:
This brings the config file back in Sync (Nearly) with what is on the server currently, less any current patches that TFM does not support.

### DIFF
--- a/src/main/resources/config.yml
+++ b/src/main/resources/config.yml
@@ -93,58 +93,89 @@ allow:
 #   - 'n:b:/ban:_'
 #
 blocked_commands:
-    # Disabled commands
+  # Disabled commands
   - 'n:b:/time:Server-side time changing is disabled. Please use /ptime to set your own personal time.'
-  - 'n:b:/gamemode:Use /creative and /survival to set your gamemode.'
+  - 'n:b:/gamemode:Use /gmc and /gms to set your gamemode.'
   - 'n:b:/ban:_'
   - 'n:b:/pardon:_'
   - 'n:b:/ban-ip:_'
+  - 'n:b:/viaversion debug:_'
   - 'n:b:/pardon-ip:_'
   - 'n:b:/toggledownfall:_'
   - 'n:b:/effect:Please use /potion to set effects.'
-  - 'n:b:/enderchest:_'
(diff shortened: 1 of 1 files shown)


src/main/resources/config.yml
```diff
@@ -93,58 +93,89 @@ allow:
...
-  - 'n:b:/pp hugeexplosion:You may not use this particle! Please use another.'
-  - 'n:b:/pp barrier:You may not use this particle! Please use another.'
+  - 'n:b:/spawnmob:_'
   - 'n:b:/setjail:_'
-  - 'n:b:/tfbanlist:Use /banlist to see and wipe the ban list'
-  - 'n:b:/tfipbanlist:Use /banlist to see and wipe the ban list'
 
   # Superadmin commands
   - 's:b:/save-all:_'
+  - 's:b:/summon:Do to security reasons, the use of /summon has been disabled.'
```
ItzLevvie: This should be ```Due to ...``` instead of ```Do to ...``` 

The same thing also goes for line 129

Wild1145: This is simply the current config on the server. I will advise Mark that the corrections are needed, once he makes the change I will update this PR with the change.

mayokek:
@Wild1145, this PR is conflicting. Might wanna fix that.

JeromSar:
@mayokek It might not be, this PR does not target the development branch.

mayokek:
@JeromSar oh, okay.

JeromSar (approved):

src/main/resources/config.yml
```diff
@@ -332,7 +370,8 @@ autokick:
 overlord_ips:
   - 176.56.237.227
   - 192.184.86.138
-  - 107.191.101.27
+  - 151.50.73.196
```
JeromSar, line 373: These should not be synced with the server. They should remain the config default

JeromSar (changes requested):

Wild1145:
This was what was sent to me, I am no longer in a position to get the latest version of this now anyway, so would advise this being re-synced with a more up to date version given how old this PR now is.

PR #2039, opened 2017-05-31
Increased text limit - Resolves #1941

### LINKED ISSUE #1941: Chat Limit for TFM 5.1 1.11
When you updated you forgot to change the chat limit from 100 to 256. As in all previous Minecraft versions leading up to 1.11 the chat limit was 100, but in 1.11 it was changed to 256. In all previous TFM versions, you have prevented chat overflow by limiting the number of characters that can be expressed in chat to the client limit, but now since you have updated the TFM to 1.11 you need to update this limit to fit with the new client threshold of 256 characters.

Wild1145:
Resolves #1941

### DIFF
--- a/src/main/java/me/totalfreedom/totalfreedommod/ChatManager.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/ChatManager.java
@@ -51,10 +51,10 @@ public class ChatManager extends FreedomService
         // Strip color from messages
         message = ChatColor.stripColor(message);
 
-        // Truncate messages that are too long - 250 characters is vanilla client max
-        if (message.length() > 250)
+        // Truncate messages that are too long - 256 characters is vanilla client max
+        if (message.length() > 256)
         {
-            message = message.substring(0, 250);
+            message = message.substring(0, 256);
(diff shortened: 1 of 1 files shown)


src/main/java/me/totalfreedom/totalfreedommod/ChatManager.java
```diff
@@ -51,10 +51,10 @@ private void handleChatEvent(AsyncPlayerChatEvent event)
         // Strip color from messages
         message = ChatColor.stripColor(message);
 
-        // Truncate messages that are too long - 100 characters is vanilla client max
-        if (message.length() > 100)
+        // Truncate messages that are too long - 250 characters is vanilla client max
+        if (message.length() > 250)
```
JeromSar: As per [the Minecraft wiki page](http://minecraft.gamepedia.com/1.11), the maximum length should be 256.

JeromSar (changes requested):

Wild1145:
@JeromSar Was not aware, the original issue has been updated and I'll update the PR shortly.

JeromSar (approved):

PR #2041, opened 2017-05-31
Added a catch for if a player is caged and kicked. Resolves #1789

### LINKED ISSUE #1789: Banning after caging
Today, I caged someone and then I banned them a minute after and they left the server but the cage was still there and I was forced to manually get rid of the glass and fix the now griefed area.

Wild1145:
Resolves #1789

### DIFF
--- a/src/main/java/me/totalfreedom/totalfreedommod/caging/Cager.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/caging/Cager.java
@@ -13,6 +13,7 @@ import org.bukkit.event.block.BlockBreakEvent;
 import org.bukkit.event.player.PlayerJoinEvent;
 import org.bukkit.event.player.PlayerMoveEvent;
 import org.bukkit.event.player.PlayerQuitEvent;
+import org.bukkit.event.player.PlayerKickEvent;
 
 public class Cager extends FreedomService
 {
@@ -85,6 +86,18 @@ public class Cager extends FreedomService
 
     @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
     public void onPlayerQuit(PlayerQuitEvent event)
+    {
+        FPlayer player = plugin.pl.getPlayer(event.getPlayer());
+        CageData cage = player.getCageData();
+
+        if (cage.isCaged())
+        {
+            cage.playerQuit();
+        }
+    }
+
+    @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+    public void onPlayerKick(PlayerKickEvent event)
     {
         FPlayer player = plugin.pl.getPlayer(event.getPlayer());
         CageData cage = player.getCageData();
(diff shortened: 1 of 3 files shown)


src/main/java/me/totalfreedom/totalfreedommod/command/Command_gtfo.java
```diff
@@ -1,6 +1,8 @@
 package me.totalfreedom.totalfreedommod.command;
 
 import me.totalfreedom.totalfreedommod.banning.Ban;
+import me.totalfreedom.totalfreedommod.caging.CageData;
+import me.totalfreedom.totalfreedommod.player.FPlayer;
 import me.totalfreedom.totalfreedommod.rank.Rank;
```
OxLemonxO: I think you mean to put these imports in the other file...

src/main/java/me/totalfreedom/totalfreedommod/command/Command_gtfo.java
```diff
@@ -1,6 +1,8 @@
 package me.totalfreedom.totalfreedommod.command;
 
 import me.totalfreedom.totalfreedommod.banning.Ban;
+import me.totalfreedom.totalfreedommod.caging.CageData;
+import me.totalfreedom.totalfreedommod.player.FPlayer;
```
lemonsked: I think you meant to put these in the other file..

JeromSar (approved):

src/main/java/me/totalfreedom/totalfreedommod/caging/Cager.java
```diff
@@ -96,6 +97,18 @@ public void onPlayerQuit(PlayerQuitEvent event)
     }
 
     @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+    public void onPlayerKick(PlayerKickEvent event)
```
lemonsked, line 100: Why ignoreCancelled? If the event is cancelled, why do you need to remove the cage?

src/main/java/me/totalfreedom/totalfreedommod/command/Command_gtfo.java
JeromSar: These will need to be removed, methinks.

JeromSar (changes requested):

src/main/java/me/totalfreedom/totalfreedommod/caging/Cager.java
JeromSar: Wild's implementation is correct. If the kick event is cancelled, the player isn't kicked, so the cage data shouldn't be removed. `ignoreCancelled` causes the event handler to skip processing if the event is cancelled.

OxLemonxO: yeah but ignoreCancelled is true, causing it to ignore the event being cancelled.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_gtfo.java
Wild1145: Yes, thats where I changed implementation mid-way through. Let me fix that.

Wild1145:
@JeromSar Unused imports removed.

JeromSar (approved):

JeromSar:
Thanks for contributing! Your PR will be merged soon.

src/main/java/me/totalfreedom/totalfreedommod/caging/Cager.java
lemonsked: Oh, never mind then. I was thinking the other way around.

PR #2052, opened 2017-06-08
Adding Founder rank to TFM Front Door

ghost:
as mark requested


Wild1145:

Could you also explain what this is even doing? I cant see what this is supposed to even achieve?

PR #2060, opened 2017-06-09
Made GTFO supports coreprotect.

ghost:
Made GTFO uses coreprotect if is loaded and enabled.

### DIFF
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_gtfo.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_gtfo.java
@@ -6,6 +6,7 @@ import me.totalfreedom.totalfreedommod.util.FUtil;
 import net.pravian.aero.util.Ips;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
+import static org.bukkit.Bukkit.getServer;
 import org.bukkit.ChatColor;
 import org.bukkit.GameMode;
 import org.bukkit.Location;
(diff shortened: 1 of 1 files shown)


src/main/java/me/totalfreedom/totalfreedommod/command/Command_gtfo.java
```diff
@@ -42,18 +43,34 @@ public boolean run(CommandSender sender, Player playerSender, Command cmd, Strin
...
+            plugin.rb.rollback(player.getName());
+
         }
-        catch (NoClassDefFoundError ex)
+
+        if (getServer().getPluginManager().isPluginEnabled("CoreProtect"))
         {
+            try
+            {
+                server.dispatchCommand(sender, "co rollback t:1d r:#global u:" + player.getName());
```
Wild1145, line 67: We should not be using a dispatch command. Thats generally considered a big no-no

src/main/java/me/totalfreedom/totalfreedommod/command/Command_gtfo.java
```diff
@@ -42,18 +43,34 @@ public boolean run(CommandSender sender, Player playerSender, Command cmd, Strin
...
+            {
+            }
+
+            // Rollback
+            plugin.rb.rollback(player.getName());
+
         }
-        catch (NoClassDefFoundError ex)
+
+        if (getServer().getPluginManager().isPluginEnabled("CoreProtect"))
```
Wild1145, line 63: Can we change this to an If-Else statement please?

Wild1145 (changes requested):

src/main/java/me/totalfreedom/totalfreedommod/command/Command_gtfo.java
JeromSar: Yes, this is in the contributing guidelines. Coreprotect should get a bridge, so TFM can interact with its API.

ghost:
@Wild1145 This has been superseded by #2067

PR #2067, opened 2017-06-16
Implement a CoreProtect bridge. Resolves #2066

### LINKED ISSUE #2066: CoreProtect Integration
Currently, the official TF server is running CoreProtect. The TFM plugin on the server has some what of a bridge, but it's not very well made and has a lot of issues. I propose the idea of adding CoreProtect integration, but have CoreProtect a completely optional plugin to have as not all servers may want to use it.  Also, as the database does get rather large over time it needs to be cleared. Lastly, there should be a way to wipe only the flatlands. All of this is currently in the current TFM plugin on the server, but needs proper integration with better code.

script-head:
This is a proper CoreProtect bridge, CoreProtect is still an optional plugin as it will not load and spam tracebacks if it isn't loaded in. I've tested this extensively with multiple scenarios and it seems to work just fine.

One bug that I can not control is that if someone is if CoreProtect is trying to modify the db file while the flatlands data is being wiped, it will spam SQLExceptions saying that the database is locked. The error spam will cease to a halt when the purging is over with. Other than this, everything else is fine. 

The reason why the rollback command doesn't use the rollback and undo rollback methods is because I wanted it to be completely optional and it will be a mess if I try to switch between the two. While the rollback method is being used, the undo rollback one isn't but I left it inside for future purposes.


pom.xml
```diff
@@ -104,6 +104,14 @@
             <scope>system</scope>
             <systemPath>${project.basedir}/lib/LibsDisguises-9.0.9.jar</systemPath>
         </dependency>
+
```
Wild1145: There must be a repo with Core Protect in... Please try and find an official repo for Core Protect and use that, adding more local files to the POM just digs a deeper hole for us to try and climb out of.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_gtfo.java
```diff
@@ -54,6 +54,9 @@ public boolean run(CommandSender sender, Player playerSender, Command cmd, Strin
         // Rollback
         plugin.rb.rollback(player.getName());
 
+        // Rollback with CoreProtect
```
Wild1145, line 58: Does this account for ASyncWorldEdit. Also see line 55, as apparently this is already done there.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_wipecoreprotectdata.java
```diff
@@ -0,0 +1,31 @@
+package me.totalfreedom.totalfreedommod.command;
+
+import me.totalfreedom.totalfreedommod.rank.Rank;
+import me.totalfreedom.totalfreedommod.util.FUtil;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+import org.bukkit.scheduler.BukkitRunnable;
+
+@CommandPermissions(level = Rank.TELNET_ADMIN, source = SourceType.ONLY_CONSOLE)
```
Wild1145, line 10: Not sure I like the idea of Telnet admins being able to wipe the database, really it should be automated and if anything manual needs doing a developer / Sys-Admin should manually wipe things themselves.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_wipeflatlands.java
```diff
@@ -24,7 +25,21 @@ public boolean run(final CommandSender sender, Player playerSender, Command cmd,
             player.kickPlayer("Server is going offline for flatlands wipe, come back in a few minutes.");
         }
 
-        server.shutdown();
+        if (!plugin.cpb.isEnabled())
```
Wild1145, line 28: Would probably suggest adding a config value and then checking before getting to this as to if the server has Core Protect bridge enabled. If so you wouldnt need to do this check as you should do it on server startup.

src/main/java/me/totalfreedom/totalfreedommod/config/ConfigEntry.java
```diff
@@ -41,6 +41,9 @@
     ADMINLIST_CLEAN_THESHOLD_HOURS(Integer.class, "adminlist.clean_threshold_hours"),
     ADMINLIST_CONSOLE_IS_SENIOR(Boolean.class, "adminlist.console_is_senior"),
     //
+    COREPROTECT_AUTO_WIPING_ENABLED(Boolean.class, "coreprotect.auto_wipe"),
+    COREPROTECT_FILE_LIMIT(Integer.class, "coreprotect.file_limit"),
+    //
```
Wild1145, line 46: As previously noted, would also suggest adding an option (Maybe further up the config?) to set if CoreProtect will even be used or not on the server.

src/main/java/me/totalfreedom/totalfreedommod/bridge/CoreProtectBridge.java
```diff
@@ -0,0 +1,241 @@
...
+
+        if (!ConfigEntry.COREPROTECT_AUTO_WIPING_ENABLED.getBoolean() || getCoreProtect() == null)
+        {
+            return;
+        }
+
+        wiper = new BukkitRunnable()
+        {
+            @Override
+            public void run()
```
Wild1145: Why is all of this within the onStart section? This should probably be split out and then called here instead of being located here...?

Wild1145 (changes requested):

pom.xml
lemonsked: doesn't exist since CoreProtect is closed source. the only downloads they give are the jars on dev.bukkit.org and spigotmc. no maven repos either.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_gtfo.java
lemonsked: yes- you need BlocksHub along with CoreProtect though, but yes it works. BlocksHub is so that CoreProtect logs AsyncWorldEdit actions.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_wipecoreprotectdata.java
lemonsked: same as wipeflatlands- that's why I guess, and telnet clans wiping maps will need to do this for flatlands. i'd prefer if there was a check if the world is adminworld, only devs and owners. if other worlds, sure allow telnets.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_wipeflatlands.java
script-head: If I'm reading this correctly, you want me to add a config value to state rather or not to use the bridge even if CoreProtect is installed and enabled?

src/main/java/me/totalfreedom/totalfreedommod/command/Command_wipecoreprotectdata.java
Wild1145: From my view, there should be no need for any manual interaction with the database except if something massively fucked up, and at that point it should be a sys-admin that fixes it / developer anyway?

src/main/java/me/totalfreedom/totalfreedommod/command/Command_wipeflatlands.java
Wild1145: Let me have another think about that, Im not convinced my original logic makes sense thinking about it. I'll let you know.

pom.xml
Wild1145: Very much less keen then for us to work with the plugin if we cannot find a way of automating the Maven integration, as it'll basically push back weeks of work elsewhere in the plugin where we've tried to automate these things ready for CI...

ghost:
This looks like a good upgrade from what is running on the server, once the required changes are done I am happy.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_wipecoreprotectdata.java
script-head: @Wild1145 In your opinion, would you rather have me make it only executable by developers, or should I just remove the command all together?

Wild1145: Providing everything else is working as expected, I dont think we need the command, server owners and developers with privileged user access will have access to the DB its self, so could run SQL queries and statements to wipe select parts of the database, though I would like a second opinion from @JeromSar or @StevenLawson on this.

script-head:
I've implemented the requested changes however I'm still waiting for the response from Madgeek or Prozza regarding the wipecoreprotectdata command.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_gtfo.java
```diff
@@ -54,6 +54,9 @@ public boolean run(CommandSender sender, Player playerSender, Command cmd, Strin
         // Rollback
         plugin.rb.rollback(player.getName());
```
ghost, line 56: why not making if there is coreprotect it skips the TFM rollback?

ghost (changes requested):

src/main/java/me/totalfreedom/totalfreedommod/command/Command_gtfo.java
script-head: Because I still want to have the rollback system on TFM to be put to use. Because in the rare event the CoreProtect data was wiped or didn't log it for some strange reason and the server hasn't been restarted, it will be rolled back by TFM. This is extreamly rare but it can happen. Also, again, I still want to make use of the TFM rollback system even if the plugin is loaded or not.  I have the rollback and undo rollback methods present if in the event they are needed.

mayokek: Maybe check if CoreProtect is enabled and use that method otherwise use TFM rollback method.

Wild1145:
So it looks like I've managed to push changes to this without realising. They shouldnt impact things once my PR is merged, it resolves the conflicts with the current Maven TravisCI PR. Apologies.

Wild1145:
The CLA has been signed, there is an issue with one of my commits where I did a stupid so it is one of my accidental commits which has not signed the CLA.

Issue #2082, opened 2017-07-16
Let non-ops do /opme


savnithyt:
Let non-ops be able to do /opme when they are deop, http://totalfreedom.boards.net/thread/53382/meop-plugin

ghost:
This has been denied in various suggestions on the forums, it is unlikely to be added.

Wild1145:
There is another issue raised that outlines the potential implementations of this, this issue does not cover one of those implementations. 

This issue has been superseded by #1764

PR #2083, opened 2017-07-16
Let's non-ops do /opme

savnithyt:
Look at: https://github.com/TotalFreedom/TotalFreedomMod/issues/2082

### DIFF
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_opme.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_opme.java
@@ -6,7 +6,7 @@ import org.bukkit.command.Command;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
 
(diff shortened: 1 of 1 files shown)


falceso:
Honestly I think someone rejected something like this for if someone gets smite they reop themselves instantly etc.

aggelosQQ:
I'll agree to this if there is something that will prevent recently smitten players or users that are marked as imposters from opping themselves.

ghost:
Same as I said on the issue: This has been denied in various suggestions on the forums, it is unlikely to be added. There are no checks for ops that have been smitten/purposefully deopped.

LiamTownsley (approved):
I agree with this. Mainly because if they should not have OP they should not be on the server and should be banned.

Wild1145:
Thank you for contributing to the project. Please read the [Contributing Guidelines](https://github.com/TotalFreedom/TotalFreedomMod/wiki/Contributing) carefully before contributing to the project in the future!

PR #2084, opened 2017-07-17
Commodore64x - /glist improvement

ghost:
Completely restructure command - removed unneeded or unused code as well as adding in glist banip, unbanip (supports range and normal ips), nameban and unbanname.

### DIFF
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_glist.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_glist.java
@@ -12,104 +12,164 @@ import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
 
 @CommandPermissions(level = Rank.SUPER_ADMIN, source = SourceType.BOTH, blockHostConsole = true)
-@CommandParameters(description = "Bans or unbans any player, even those who are not logged in anymore.", usage = "/<command> <purge | ban <username> [reason] | unban <username>>")
+@CommandParameters(description = "Bans or unbans any player, even those who are not logged in anymore.", usage = "/<command> <purge | ban <username> [reason] | unban <username> | banip <ip> <reason> | unbanip <ip> | nameban <name> | unbanname <name>>")
 public class Command_glist extends FreedomCommand
 {
 
     @Override
     public boolean run(CommandSender sender, Player playerSender, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
     {
-        if (args.length < 1)
-        {
-            return false;
-        }
-
-        if (args.length == 1)
-        {
-            if ("purge".equals(args[0]))
-            {
-                checkRank(Rank.SENIOR_ADMIN);
-                plugin.pl.purgeAllData();
-                msg("Purged playerbase.");
-
-                return true;
-            }
-
-            return false;
-        }
-
         if (args.length < 2)
         {
             return false;
         }
 
-        String username;
+        String username = null;
         final List<String> ips = new ArrayList<>();
-
+        boolean usingIp = false;
+        String banIp = null;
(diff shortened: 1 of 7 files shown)


src/main/java/me/totalfreedom/totalfreedommod/command/Command_glist.java
```diff
@@ -12,104 +12,164 @@
...
 
     @Override
     public boolean run(CommandSender sender, Player playerSender, Command cmd, String commandLabel, String[] args, boolean senderIsConsole)
     {
-        if (args.length < 1)
-        {
-            return false;
-        }
-
-        if (args.length == 1)
```
Wild1145, line 27: You've removed this however still listed it as a potential argument.

Wild1145 (changes requested):

ghost:
@Wild1145 All issues have been rectified

lemonsked:
Why  remove purge?

Wild1145:

src/main/java/me/totalfreedom/totalfreedommod/command/Command_glist.java
```diff
@@ -12,104 +12,164 @@
...
+                if (player != null)
+                {
+                    player.kickPlayer(playerBan.bakeKickMessage());
+                }
+                return true;
+            case "unban":
+            case "pardon":
+                if (usingIp)
+                {
+                    msg("Please specify a player, not an ip.");
```
LiamTownsley, line 88: You could make it so you can /glist unban [USERNAME/IP]

mayokek:
`/glist purge` is actually useful as `/gtfo` isn't made for 24-48 hours ban, it permanent ban them.

Wild1145 (changes requested):
The functionality that purge offers is something that we would like to remain in the project.

ghost:
@Wild1145 Purge is not really ever used and causes issues since it doesn't just purge ban data.

Wild1145:
@Commodore64x I understand it is fairly regularly used, and if it is breaking /cage when we need to fix /cage. I thought I had done that with a recent PR for cage improvements anyway but that might not be the case.

ghost:
@Wild1145 Searching logs from the last 10 days shows noone has used it once, it really isn't practical to have a command that clears all fplayer and playerdata data as it does break a lot of things.

Wild1145:
@Commodore64x Can you elaborate on what exactly it breaks?

src/main/java/me/totalfreedom/totalfreedommod/command/Command_glist.java
```diff
@@ -12,104 +12,164 @@
...
+                plugin.bm.addBan(nameBan);
 
-            for (String ip : ips)
-            {
-                Ban ban = plugin.bm.getByIp(ip);
-                if (ban != null)
+                if (player != null)
                 {
-                    plugin.bm.removeBan(ban);
+                    player.kickPlayer(nameBan.bakeKickMessage());
```
LiamTownsley, line 123: `.bakeKicKMessage()`?

LiamTownsley (changes requested):

src/main/java/me/totalfreedom/totalfreedommod/command/Command_glist.java
lemonsked: It's the same as before.

ghost:
@Wild1145 It simply isn't ever used. It clears fPlayer and player data which is far from ideal. Things like muting, blocked commands, cages, freezing has issues and the /purgeall command is favored as a safer and easier way to achieve the unfreezing/caging etc of players.

mayokek:
And how does the banning affect the player data?

ghost:
@mayokek It doesn't, however /glist purge does.

mayokek:
@Commodore64x, `/glist purge` does only purge banlist...

mayokek:
@Commodore64x, I see the issue, it purges the player data, not banlist. Might wanna change `plugin.pl.purge();` to `plugin.bm.purge();` if you are going to add it back.

Wild1145:
@Commodore64x We do not want to be removing functionality for shitts and giggles just because you dont see a use case for it. The issue stands that you are simply removing functionality because you feel like it, without any associated issue for discussion. Until an approved issue is raised to remove the /glist purge functionality, it is to remain in TFM.

PR #2091, opened 2017-08-01
Update

CharlesYT69:
From the last one I put human but I changed it to player.


lemonsked:
P in Player doesn't need to be capitalized.
I hate it when instead of clicking comment I click close and comment.

StevenLawson:
I'd rather keep this like it is. Sure, it might not be the most politically correct thing to say, but I like the humor in it.

PR #2096, opened 2017-08-09
Create Command_savebuild.java

### LINKED ISSUE #2095: A command to show ops how to save builds with worldedit
We need a command to help ops with saving with world edit.

hitzzaf:
fixes #2095


mayokek:
According to contributing guidelines, we would like to avoid a `wall-of-text-`.

StevenLawson:
Commands that display a hard coded "wall-of-text", and do nothing else, are frowned upon in TotalFreedomMod. They're usually only applicable to a specific server, and you have to recompile the plugin to change any of the text or formatting. A forum post with images or video that showed this specific WorldEdit saving process would be much more helpful.

That being said, a much more flexible command for info-dumps would probably be well received. Allow the server operator configure a YAML file with multiple command triggers, with each command trigger having a formatted body of text.

@hitzzaf - Sorry, but I can't accept a basic "wall-of-text" command.

untuned:
If I were making the command, I'd link to a forum post explaining how to do it instead.  This would prevent the "Wall of text" issue that @StevenLawson said.

Wild1145:
@Untuned It does however still require TFM to be re-compiled for every server.

PR #2099, opened 2017-08-11
CustomItems

shrimpuu:
I've integrated my new items into TFM, they are ready to be added. Thanks!

View my main thread here: http://totalfreedom.boards.net/thread/53452/custom-items-plugin-complete

also, there are effects, but mostly still in progress.

### DIFF
--- a/command/Command_excalibur.java
+++ b/command/Command_excalibur.java
@@ -1,0 +1,45 @@
+import java.util.ArrayList;
+import java.util.List;
+import org.bukkit.ChatColor;
+import org.bukkit.Material;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandExecutor;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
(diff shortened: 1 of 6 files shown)


StevenLawson:
Please review our [contributing guidelines](https://github.com/TotalFreedom/TotalFreedomMod/wiki/Contributing).

These commands would need to be actually be integrated into TFM before we could even consider pulling them in. Right now, they wouldn't even be loaded since they're in the wrong package and they don't extend FreedomCommand.

I'm not sure it's necessary to make six different commands that only give the user themed items. A single command with a more flexible item customization scheme (possibly backed by a YAML file) would be much more useful.

shrimpuu:
Madgeek- I am in the workings of creating effects for all of the items, and working out my errors, I completely forgot to extend as a freedomcommand, as i didnt code them with TFM and tried my best to charge it. I honestly don't know why I even opened this up, I will recreate this later when i get everything figured out.

PR #2100, opened 2017-08-12
Add WipeWorldGuard Regions

ghost:
There is too much regions?

You want to wipe every regions because OPS doesn't have access to build on it?

Just use this command

Coded by me and @lemon for TFM


lemonsked:
Probably should add a Maven dependency for WorldGuard. @marcocorriero

ghost:
@OxLemonxO #2101

mayokek:
#2101 can easily be added here.

PR #2104, opened 2017-08-13
Upgrade Command_gcmd and Command_wildcard

Mafrans:
Upgrades Command_gcmd and Command_wildcard to use a variable system allowing the sender to use the commands in more ways.

Variables added:
- ?name, displays the player's name
- ?nick, displays the player's current nickname
- ?rank, displays the player's current rank
- ?crank, displays the player's current rank with color coding
- ?title, displays the player's current title, identical with ?rank unless you are either dev, owner or founder
- ?ctitle, displays the player's current title with color coding
- ?world, displays the name of the world the player is currently in
- ?x | y | z, displays the player's x, y or z variable as a decimal point float
- ?ix | iy | iz, displays the player's x, y or z variable as a real integer
- ?pitch | yaw, displays the player's rotation along the pitch or yaw axes respectively

Do note that at this point the variables are CASE SENSITIVE (as in, a variable is required to use lower case letters) and there is currently no way to escape a variable, which means you will have to use either upper case letters (as the variables are lower case specific) or another way to override the variable, such as writing them as "?.<variable>". Both of these issues will likely be fixed later when I have the time necessary.

### DIFF
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_gcmd.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_gcmd.java
@@ -2,9 +2,12 @@ package me.totalfreedom.totalfreedommod.command;
 
 import me.totalfreedom.totalfreedommod.rank.Rank;
 import org.apache.commons.lang3.StringUtils;
+import org.bukkit.ChatColor;
 import org.bukkit.command.Command;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
+
+import java.util.HashMap;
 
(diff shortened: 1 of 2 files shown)


Wild1145:

ghost:
Very nice!

Mafrans:
This was designed specifically as an addition, and does not resolve an actual issue created on github.

Wild1145:
@Mafrans Without an associated issue this cant be merged. See the Contributing Guidelines.

Mafrans:
As there is no issue connected to this addition, and i will not be creating one, this is now closed.

Wild1145:
@Mafrans As you wish.

PR #2116, opened 2017-08-29
A Copy of Cake- but proposed as Potato

mandar-ghade:
A Copy of Cake- but proposed as Potato.


ncurran02:
First of all, this is copied right from `Command_cake` class, which is against the guidelines of copying.
Secondly, you're missing the `.java` extension in the filename and this haven't been tested.
And lastly, there are no issues to link to this PR.

mandar-ghade:
Requesting Pull Request to be closed.

Issue #2125, opened 2017-10-14
./tree and /bigtree


DeternoMC:
Both commands are used for griefing; upon ban doesn't roll back any created blocks.

Gusted:
Why blocking?

Gusted:
a reason please

Gusted:
issue close!

JeromSar:
Perhaps we can figure out a way to roll these back. @Wild1145 any ideas?

Gusted:
Admins can rollback it

Issue close

Wild1145:
@Gusted /ro would risk removing more than the tree that was just placed. 

The issue here is that //undo does not appear to roll it back, though I am not entirely sure why, as I understand it, the command simply creates a tree / big tree at the location you are looking at? Not sure why //undo would not work with it...

aokod:
Use CoreProtect. `/rollback` is very outdated.

Wild1145:
@ir1x TFM does not officially support CoreProtect at the current time. The TFM version of Rollback or TF-WorldEdit are the plugins a solution should be found through. As I mentioned, WorldEdit does not appear to track these commands and that is an issue.

PR #2126, opened 2017-10-14
Block /tree and /bigtree

DeternoMC:
Issue: #2125


lemonsked:
I don't think we need to block the commands on the official TFM repo, make a suggestion on TF to block it.

Wild1145:
We try to periodically sync our config with the official TF server config along with basic sanity checking. We wont add anything the official server does not have and base our default config off of their configuration files.

PR #2132, opened 2017-12-02
Make /say More Noticeable?

FoxyCodes:
Hey?. I am NoobsAreGod and I tweaked the /say command to make it alittle more noticeable as since it's a lighter color I think it is less noticed, Here is the source code of what I tweaked

https://pastebin.com/yf22ZRUE

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # TotalFreedomMod #
 
-[![Build Status](https://travis-ci.org/Wild1145/TotalFreedomMod.svg?branch=maventravisci)](https://travis-ci.org/Wild1145/TotalFreedomMod)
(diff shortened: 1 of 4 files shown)


ncurran02:
To be honest, I like it how it is. And also, if you want to raise a suggestion, please put it in the issue section.

LiamTownsley (changes requested):
This does not have an "issue" and the code changed in the PR does not match the description of the PR...

robingall2910 (changes requested):
redo this entire thing like holy crap fam

lemonsked:
This isn’t a proper PR. Closed.

Issue #2121, opened 2017-09-14
/cake broken in 1.12 prerelease


lemonsked:
When you do /cake, it says ingame: Command error: not supported in this minecraft version. 
This is due to the fact that achievements do not exist in versions of minecraft later then 1.12
Replaces with advancements.
Error in logs:

```:[14:45:20 ERROR]: [TotalFreedomMod] Uncaught exception executing command: cake
:[14:45:20 ERROR]: [TotalFreedomMod] nulljava.lang.UnsupportedOperationException: Not supported in this Minecraft version.
	at org.bukkit.craftbukkit.v1_12_R1.entity.CraftPlayer.awardAchievement(CraftPlayer.java:593)
	at me.totalfreedom.totalfreedommod.command.Command_cake.run(Command_cake.java:47)
	at me.totalfreedom.totalfreedommod.command.FreedomCommand.runCommand(FreedomCommand.java:53)
	at me.totalfreedom.totalfreedommod.command.FreedomCommandExecutor.onCommand(FreedomCommandExecutor.java:81)
	at org.bukkit.command.PluginCommand.execute(PluginCommand.java:44)
	at org.bukkit.command.SimpleCommandMap.dispatch(SimpleCommandMap.java:141)
	at org.bukkit.craftbukkit.v1_12_R1.CraftServer.dispatchCommand(CraftServer.java:651)
	at net.minecraft.server.v1_12_R1.PlayerConnection.handleCommand(PlayerConnection.java:1386)
	at net.minecraft.server.v1_12_R1.PlayerConnection.a(PlayerConnection.java:1221)
	at net.minecraft.server.v1_12_R1.PacketPlayInChat.a(PacketPlayInChat.java:45)
	at net.minecraft.server.v1_12_R1.PacketPlayInChat.a(PacketPlayInChat.java:1)
	at net.minecraft.server.v1_12_R1.PlayerConnectionUtils$1.run(SourceFile:13)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at net.minecraft.server.v1_12_R1.SystemUtils.a(SourceFile:46)
	at net.minecraft.server.v1_12_R1.MinecraftServer.D(MinecraftServer.java:748)
	at net.minecraft.server.v1_12_R1.DedicatedServer.D(DedicatedServer.java:406)
[...]

PR #2133, opened 2017-12-05
Fixes the error with /cake

xensace:
https://github.com/TotalFreedom/TotalFreedomMod/issues/2121


Wild1145:
I dont think this is a proper solution at all... Advancements have replaced Achievements and we should be able to port functionality accordingly.

Issue #2115, opened 2017-08-21
/tag clearall fix


hitzzaf:
Make /tag clearall only clear non admins tag's

Wild1145:
I think maybe /tag clearall should reset everyones tags back to their default tag (Not remove them). I would still like the option of resetting admins tags, as I have seen some of the stupid ones they've done, but yes by default perhapse it does just need to be OPS.

PR #2134, opened 2017-12-05
Fixes /tag clearall

xensace:
https://github.com/TotalFreedom/TotalFreedomMod/issues/2115

### DIFF
--- a/src/main/java/me/totalfreedom/totalfreedommod/command/Command_tag.java
+++ b/src/main/java/me/totalfreedom/totalfreedommod/command/Command_tag.java
@@ -54,11 +54,19 @@ public class Command_tag extends FreedomCommand
                 int count = 0;
                 for (final Player player : server.getOnlinePlayers())
                 {
-                    final FPlayer playerdata = plugin.pl.getPlayer(player);
-                    if (playerdata.getTag() != null)
+                    if (!plugin.al.isAdmin(sender))
                     {
-                        count++;
-                        playerdata.setTag(null);
+                        final FPlayer playerdata = plugin.pl.getPlayer(player);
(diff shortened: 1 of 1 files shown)


ncurran02:
That doesn't solve the issue.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_tag.java
```diff
@@ -54,11 +54,19 @@ else if ("clearall".equalsIgnoreCase(args[0]))
...
                 {
-                    final FPlayer playerdata = plugin.pl.getPlayer(player);
-                    if (playerdata.getTag() != null)
+                    if (!plugin.al.isAdmin(sender))
+                    {
+                        final FPlayer playerdata = plugin.pl.getPlayer(player);
+                        if (playerdata.getTag() != null)
+                        {
+                            count++;
+                            playerdata.setTag(null);
```
Wild1145, line 63: Setting a tag to null sounds like a really really bad idea...

src/main/java/me/totalfreedom/totalfreedommod/command/Command_tag.java
```diff
@@ -54,11 +54,19 @@ else if ("clearall".equalsIgnoreCase(args[0]))
...
+                        }
+                    }
+                    else
                     {
-                        count++;
-                        playerdata.setTag(null);
+                        // Give the admin their default tag
```
Wild1145, line 68: You never increment the counter, you probably want a player and admin counter if you were to do something like this.

Wild1145 (changes requested):

PR #2152, opened 2018-01-25
Revert "removed /spectator "

mgflames:


mgflames:
Accidently posted - ignore.

Issue #1652, opened 2016-05-12
/ov and /overlord alias mis-match


Telesphoreo:
When renaming Command_overlord.java to Command_ov.java you didn't change the alias to /overlord

Also in /mobspawn the description is "Make an announcement"

PR #2163, opened 2018-05-24
TFM 5.0.1 1.12 Release

Wild1145:
So far the testing has been positive and no issues have been identified. One issue I've noticed in the PR that we may need to address as a wider issue is around the ReadME as the TravisCI link does not match as its branch specific. 

I think I've also potentially confused the version increments, this was supposed to be a 0.0.1 increment on the current release but I think I've mixed 5.0 and 5.1 up in a few places now...


Wild1145:
@OxLemonxO Are you able to sign the CLA please? IT appears that when you accidentally pushed to this branch it kinda broke things...

pom.xml
```diff
@@ -210,26 +237,66 @@
...
+                    <prefix>git</prefix>
+                    <dateFormat>yyyy-MM-dd HH:mm:ss</dateFormat>
+                    <verbose>false</verbose>
+                    <generateGitPropertiesFile>true</generateGitPropertiesFile>
+                    <generateGitPropertiesFilename>${basedir}/src/main/resources/git.properties</generateGitPropertiesFilename>
+                    <format>properties</format>
+                    <failOnNoGitDirectory>true</failOnNoGitDirectory>
+                    <failOnUnableToExtractRepoInfo>false</failOnUnableToExtractRepoInfo>
+                    <excludeProperties>
+                        <excludeProperty>git.branch</excludeProperty>
```
JeromSar: Considering we also have an `includeOnlyProperties`, is this list not redundant. It seems to me like there must be a better way to only include the right properties without excluding every other property explicitly.

src/main/java/me/totalfreedom/totalfreedommod/TotalFreedomMod.java
```diff
@@ -259,22 +259,31 @@ public void load(TotalFreedomMod plugin)
             try
             {
                 final Properties props;
+                final Properties gitprops;
                 try (InputStream in = plugin.getResource("build.properties"))
                 {
                     props = new Properties();
                     props.load(in);
                 }
+                try (InputStream in = plugin.getResource("git.properties"))
```
JeromSar, line 268: Is there any way where we can include the git properties in the build properties? From a logical perspective, git properties signify the build process/state and it would make sense to centralize this data in one file.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_cake.java
```diff
@@ -43,12 +42,11 @@ public boolean run(CommandSender sender, Player playerSender, Command cmd, Strin
...
-            player.awardAchievement(Achievement.BAKE_CAKE);
+            
         }
 
         FUtil.bcastMsg(output.toString());
 
         return true;
     }
-}
+}
```
JeromSar: Should have a newline.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_ov.java
```diff
@@ -15,7 +15,7 @@
 import org.bukkit.entity.Player;
 
 @CommandPermissions(level = Rank.IMPOSTOR, source = SourceType.ONLY_IN_GAME)
-@CommandParameters(description = "Overlord - control this server in-game", usage = "access", aliases = "ov")
+@CommandParameters(description = "Overlord - control this server in-game", usage = "access", aliases = "overlord")
```
JeromSar, line 18: Perhaps these should be the other way around. The full filename should be `Command_overlord.java`, and the alias `ov`.

src/main/resources/config.yml
```diff
@@ -11,15 +11,15 @@ server:
...
+  # Shown at the server list - Please note that support for %mcversion% has been withdrawn
+  motd: TotalFreedom &8- The Original All-Op Server!
 
   # Every word in the MOTD will be a different color unless prefixed by a color code
   colorful_motd: true
 
   # All players who show up as owner
   owners:
-    - markbyron
+    - TheMinecraft
```
JeromSar, line 22: Unless you know who 'TheMinecraft' is, we might as well leave this an empty array.

pom.xml
```diff
@@ -4,18 +4,19 @@
 
     <groupId>me.totalfreedom</groupId>
     <artifactId>totalfreedom</artifactId>
-    <version>5.0</version>
+    <version>5.1</version>
```
JeromSar: Considering this is still TFM Electrum, and not TFM _F_. We should stay at `5.0`.

pom.xml
```diff
@@ -140,23 +137,51 @@
...
                 <executions>
                     <execution>
                         <phase>initialize</phase>
                         <configuration>
                             <target>
-                                <propertyfile file="build.properties" comment="Build information. Edit this to your liking.">
+                                <propertyfile file="src/main/resources/build.properties" comment="Build information. Edit this to your liking.">
                                     <entry key="buildAuthor" default="unknown" />
+                                    <entry key="buildNumber" default="0" />
+                                    <entry key="buildCodeName" default="Electrum (MC 1.12) Pre-Release 1" />
```
JeromSar: Is there any way not to hardcode the code name in the pom file? Could we instead hard code it in build.properties?

JeromSar (changes requested):
Overall, good job!

src/main/resources/config.yml
ncurran02: TheMinecraft is Windows aka the owner of TF

pom.xml
Wild1145: These are the build defaults for the POM. You can change them when the build.properties are generated but I wanted to ensure there are always default values as part of the generation.

pom.xml
Wild1145: Yes, I think I got some of the versioning confused over the last few months. Shall fix this.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_ov.java
Wild1145: So this was just to fix a bug where the alias and file name were exactly the same. If you think we should change it I am inclined to say we should raise a new issue for the change and put it in a later release.

src/main/resources/config.yml
Wild1145: Exactly what @Focusvity has said.

src/main/java/me/totalfreedom/totalfreedommod/TotalFreedomMod.java
Wild1145: I did try to do this, but there was no easy way I could find (After 2 days of research) to do this without either not updating the git properties, or wiping the other build properties and wiping them back to their defaults. I'd like to have a better solution as this is certainly not perfect, but I couldn't come up with one unfortunately.

pom.xml
Wild1145: I've just done some testing and it looks like that is the case, it adds a bit more info to the git.properties file but nothing too major.

src/main/java/me/totalfreedom/totalfreedommod/command/Command_ov.java
JeromSar: If you fix something that's broken, fix it good then. From a gameplay perspective, it doesn't change much, solely the help pages.

src/main/resources/config.yml
JeromSar: okay!

src/main/java/me/totalfreedom/totalfreedommod/command/Command_ov.java
Wild1145: The ticket requirements clearly state that the alias should be changed to /overlord from /ov as per the issue https://github.com/TotalFreedom/TotalFreedomMod/issues/1652 - That is what I have done.

JeromSar:
Will look at this ASAP.

Telesphoreo:
I figured out how to merge the git properties into the build properties. You can see how I've done it here:
pom.xml: https://github.com/Telesphoreo/TotalFreedomMod/blob/1.13/pom.xml
TotalFreedomMod.java: https://github.com/Telesphoreo/TotalFreedomMod/blob/1.13/src/main/java/me/totalfreedom/totalfreedommod/TotalFreedomMod.java

Wild1145:
Thanks @Telesphoreo - I'll take a look at implementing such an improvement as a separate issue for the 1.13 release.

PR #2180, opened 2019-09-22
Oraclejdk8 doesn’t seem to working anymore.

ncurran02:


Wild1145:
I might well need to merge this in, I think the other change I've made to force it to use the older version of Ubuntu should fix it, currently the issue is with Essentials not managing to be found and load in, but that's a different issue.

ncurran02:
The only reason the essentials doesn’t work is because the server is returning bad gateway. You can check it for yourself: http://repo.ess3.net/content/groups/essentials

Wild1145:
I'm currently looking into this to find an alternative repo location assuming the current Essentials one is not restored. Once we have a working repo and a locally working build we can see if this is still needed, as I think I've already fixed the JDK Issue within the Travis System.

ncurran02:
Perhaps use EssentialsX instead?
The maven repo for this: https://ci.ender.zone/plugin/repository/everything/
EssentialsX location: https://ci.ender.zone/plugin/repository/everything/net/ess3/EssentialsX/

Wild1145:
Apologies I managed to miss your message back last month, I'll look into EssentialsX, hopefully it should present the same API's and such for TFM to be able to use at build time.

ncurran02:
I’m closing this PR as the issue is with Essentials not with JDK.

PR #2181, opened 2019-10-12
Change /nicknyan name length to 3

zD12:


Wild1145:
@zD12 What's the purpose of the reduction from 4 to 3 characters?

ncurran02:
You can have minimum of 3 characters in your name for Minecraft.

PinkPanther1046:
I'm pretty sure I can fix this

Wild1145:
Ahh that I did not realise. That's cool, I'll give it a once over when I get the CI working and we can merge it in assuming nothing obvious breaks (Which would be impressive)

PR #2183, opened 2019-12-22
Fix Essentials

ncurran02:
Correct repo used, the bridge should work the same.


Wild1145:
Thank you, I was trying to do this half asleep yesterday :)

